PR #347, opened 2025-02-19
Fix: reflect properties

emilygorelik:
add `reflect:true` to all properties to mimic native components

### DIFF
--- a/packages/nys-button/src/nys-button.ts
+++ b/packages/nys-button/src/nys-button.ts
@@ -7,8 +7,8 @@ import { ifDefined } from "lit/directives/if-defined.js";
 let buttonIdCounter = 0; // Counter for generating unique IDs
 
 export class NysButton extends LitElement {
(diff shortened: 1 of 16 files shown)


NovaCat35 (approved):
LGTM

packages/nys-button/src/nys-button.ts
```diff
@@ -64,8 +64,9 @@ export class NysButton extends LitElement {
       ? (value as (typeof NysButton.VALID_TYPES)[number])
       : "button";
   }
-  @property({ type: Function }) onClick: (event: Event) => void = () => {};
-  @property({ type: String }) href = "";
+  @property({ type: Function, reflect: true }) onClick: (event: Event) => void =
```
plasticmind, line 67: @emilygorelik Can you just confirm that a function can/should have its value reflected?

plasticmind (approved):
Approved.

PR #349, opened 2025-02-20
Fix: handle invalid and showError

NovaCat35:
# Summary

Make `showError` toggle when the user submits the form with unanswered fields, rather than when they interact with a field and then undo their input.

## Breaking change


plasticmind (approved):
Made a few small changes for accessibility and validation handling logic

PR #355, opened 2025-02-24
Fix: globalfooter/header link

NovaCat35:
# Summary

Removed default linking from [window.location.origin](https://developer.mozilla.org/en-US/docs/Web/API/Location/origin) to no link on the global footer's agency name.

Revise global header storybook

## Breaking change


emilygorelik (approved):
lgtm

esteinborn (approved):
LGTM

PR #357, opened 2025-02-24
Fix: globalfooter link states

NovaCat35:
# Summary

Global component and unav-footer links hover/active state is revised. 

## Breaking change

### DIFF
--- a/packages/nys-globalfooter/src/nys-globalfooter.styles.ts
+++ b/packages/nys-globalfooter/src/nys-globalfooter.styles.ts
@@ -115,6 +115,13 @@ export default css`
     letter-spacing: var(--_nys-globalfooter-link-letterspacing);
   }
 
+  a:hover {
+    text-decoration: underline;
+  }
+  a:active {
+    text-decoration-thickness: 12%;
+  }
+
   /* Specific layout for menu links (grouped or singular list of menus) */
   .nys-globalfooter__content {
     width: 100%;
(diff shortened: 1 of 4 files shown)


packages/nys-globalfooter/src/nys-globalfooter.styles.ts
```diff
@@ -115,6 +115,13 @@ export default css`
     letter-spacing: var(--_nys-globalfooter-link-letterspacing);
   }
 
+  a:hover {
+    text-decoration: underline;
+  }
+  a:active {
+    text-decoration-thickness: 12%;
```
emilygorelik, line 122: is there a reason we're doing this as a percent and not as one of our variables (ie: --nys-space-2px)

packages/nys-globalheader/src/nys-globalheader.styles.ts
```diff
@@ -176,6 +176,10 @@ export default css`
     text-underline-position: from-font;
   }
 
+  .nys-globalheader__content ul a:active {
+    text-decoration-thickness: 12%;
```
emilygorelik, line 180: same

emilygorelik (approved):
approving, but please answer my question about the text decoration thickness

packages/nys-globalfooter/src/nys-globalfooter.styles.ts
esteinborn: lets use variables wherever possible

emilygorelik (approved):

esteinborn (approved):
LGTM

PR #362, opened 2025-02-25
Fix/description slot

emilygorelik:
# Summary

Allow the description to be added to the label as a slot rather than a prop to allow for rich text and links. The feature was originally lost when replacing the labels with a component.


esteinborn (approved):
LGTM

PR #375, opened 2025-03-05
Reflect "name" prop and fix showErrrrorr on radiogroup

NovaCat35:
# Summary

Reflect "name" prop on components and fix showError issue on `nys-radiogroup`

## Breaking change

## Solution
This is to solve the issue of React demo not submitting the form.

## Feedback needed 
I remember Jesse in the prev weeks had pointed out what to reflect or not, so my question is should the "name" prop be reflected (even though it fixes the React form submission issue)?
https://lit.dev/docs/components/properties/#reflected-attributes
Screenshot 2025-03-05 at 12 12 26 PM


esteinborn (approved):
LGTM

PR #380, opened 2025-03-11
Feature/plop integration

emilygorelik:
Added the ability to generate new component files through the command `npm run gen`. The developer will then provide the name of the component and it's version and it will generate the necessary files for the component to start development.


esteinborn (approved):
LGTM

PR #382, opened 2025-03-12
Fix: form issues (Error Validation)

NovaCat35:
# Summary
Fix validation on `_handleInvalid()` not to use native popup validations. Instead, prevent it and use focus() on components when constraint fails by manually going through the form and getting the first invalid component.

## Breaking change
Not a breaking change

## Concerns
This logic is implemented in the `_handleInvalid()` of all components (some logic exception for radio button).
Because we're looping the form in each invalid form component, I have some concern on the run-time and speed.

## Testing
Done in https://github.com/ITS-HCD/nysds-react-demo

### DIFF
--- a/packages/nys-toggle/src/nys-toggle.ts
+++ b/packages/nys-toggle/src/nys-toggle.ts
@@ -59,7 +59,7 @@ export class NysToggle extends LitElement {
   /********************** Form Integration **********************/
   // Update the internals whenever `checked` or `value` changes.
   updated(changedProperties: Map<string, any>) {
-    if (changedProperties.has('checked') || changedProperties.has('value')) {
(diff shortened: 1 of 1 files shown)


packages/nys-toggle/src/nys-toggle.ts
```diff
@@ -49,6 +56,15 @@ export class NysToggle extends LitElement {
     }
   }
 
+  /********************** Form Integration **********************/
+  // Update the internals whenever `checked` or `value` changes.
+  updated(changedProperties: Map<string, any>) {
+    if (changedProperties.has('checked') || changedProperties.has('value')) {
```
esteinborn, line 62: Lint fix please

esteinborn (changes requested):
Lint fix please

PR #419, opened 2025-03-24
Fix: UNav Header documentation sizing and media query refactor

emilygorelik:
There was an issue with the unav header in documentation where it was spilling out of the container. this is because it was listening to the screen size and not it's parent. modified component to listen to parent size to avoid the issue in doucmention.
before:
image
after:
image


packages/nys-unavheader/src/nys-unavheader.ts
```diff
@@ -15,14 +15,18 @@ export class NysUnavHeader extends LitElement {
   @property({ type: Boolean }) hideSearch = false;
   @property({ type: String, reflect: true })
   get screenSize(): string {
-    const width = window.innerWidth;
+    const width = this.parentContainer
```
plasticmind, line 18: @emilygorelik This would be far more performant using purely CSS, especially since the event listener on line 114 could run dozens or even hundreds of times per second during a resize. I know you're using the screenSize() check throughout the script, but you could tackle this display logic with CSS breakpoints. I proofed this out when I initially put together [this CodePen demo](https://codepen.io/plasticmind/pen/poBraQp) (style outdated, but CSS logic valid). 

Let me know if you'd like to talk through this. /cc @esteinborn

plasticmind (changes requested):
Let's approach the resize logic with CSS instead of JS.

NovaCat35 (approved):
LGTM!

packages/nys-unavheader/src/nys-unavheader.ts
```diff
@@ -101,24 +92,93 @@ export class NysUnavHeader extends LitElement {
...
-  }
-
   connectedCallback() {
     super.connectedCallback();
-    window.addEventListener("resize", this._updateScreenSize.bind(this));
+
+    // Check if the component has a parent element (should only be for documentation)
+    if (this.parentElement) {
+      this.style.containerType = "inline-size"; // Enable container query if inside a parent
+    }
```
esteinborn, line 101: Does this need to be here officially? If its only to fix the reference website usage of it, we should take it out.

esteinborn (changes requested):
Please remove the check for storybook specific fix, lets move a fix into nysds-site.js file instead of baking into the component itself

packages/nys-unavheader/src/nys-unavheader.ts
emilygorelik: @esteinborn Is this still needed a refactor or further explanation or can this be merged in?

esteinborn (approved):
LGTM

PR #431, opened 2025-03-31
make undefined components invisible

emilygorelik:
Make components invisible while JS loads to avoid flashing unstyled components on screen


packages/styles/src/nysds.css
```diff
@@ -21,5 +21,8 @@
 
 /* TODO: Include Global Styles (default typography and body styles) */
```
NovaCat35, line 22: Don't forget to remove this comment once you're done

NovaCat35 (approved):
LGTM! (see minor comment)

esteinborn (approved):
LGTM

PR #475, opened 2025-04-22
allow dev to customize target type of button when passing a URL

emilygorelik:
Add `target` prop to allow devs to customize the url target. default is "_self"


esteinborn (approved):
LGTM

PR #476, opened 2025-04-24
Fix: Radiobutton Text Alignment

emilygorelik:
Issue flagged about how text was not centered properly relative to the radio button. Issue was caused by margin bleeding into the custom radio button styles that was set in the user agent stylesheet. By changing `margin-bottom: auto` to `margin: 0 0 auto 0` though almost the same thing, properly overrides the stylesheet and fixes the misalignment. 
before: 
image
image
after:
image
image


esteinborn (approved):
LGTM

Issue #502, opened 2025-04-29
BUILD: TEST - Textinput Button Slot


NovaCat35:
### Discussed in https://github.com/ITS-HCD/nysds/discussions/390

<div type='discussions-op-text'>

<sup>Originally posted by **esteinborn** March 20, 2025</sup>
### Component name

Enhancement to the existing [`<nys-textinput>` component](https://its-hcd.github.io/nysds/?path=/docs/components-textinput--docs) component — add optional `end-action` slot

### Component description

Enhance the `nys-textinput` component to support a new optional `end-action` slot. This lets teams embed a contextually relevant button inside the input field. The button would sit flush against the input field border and be aligned to the "inline-end" — using logical CSS properties so that it works as expected with RTL languages. 

### Why the NYS Design System needs this component

Buttons embedded inside a text input is a common pattern across applications and sites, for things like search inputs, address or ZIP lookups, etc. Without this, teams would need to either build new or re-implement this pattern inconsistently or use layout workarounds.

### How the component works

**Slot Name:** "action-end" (matches CSS logical property naming)

### Real-world examples

Universal Navigation Bar:
![image](https://github.com/user-attachments/assets/71c90541-cef5-4d07-8c17-7f7ea4d0788c)

### When to use this component and when to consider something else

Using a button inside an input field can improve usability when it provides a clear action **related to the input**.  Ensure the button is intuitive and directly related to the input action.

For example:
Search Fields – A search bar with a magnifying glass icon or a “Search” button inside or at the end of the field. This makes it clear how to submit the query.

### Usability considerations

- Do not allow multiple buttons inside the same slot

### Accessibility considerations

- The button should be focusable (tabindex="0" if necessary)
- Users should be able to activate it with Enter or Space
- If the button has only an icon (e.g., a search icon), use aria-label</div>

mrhassan365:
Accessibility (A11y) Test Checklist

**1. Keyboard Navigation**
- [ ]  Tabbing lands on the input field and then the end-action button
- [ ]  Tab and Shift+Tab navigate correctly between input and button
- [ ]  Enter key on the button triggers the intended action
- [ ]  Button is reachable even when input has content

**2. Screen Reader Compatibility**
- [ ]  Input is properly labeled (visible label or aria-label)
- [ ]  Button has an accessible name (via visible text, aria-label, or aria-labelledby)
- [ ]  Button role is correctly announced
- [ ]  Screen reader announces input and button in logical order

[...]

Issue #501, opened 2025-04-29
BUILD: A11Y - Textinput Button Slot


NovaCat35:
### Discussed in https://github.com/ITS-HCD/nysds/discussions/390

<div type='discussions-op-text'>

<sup>Originally posted by **esteinborn** March 20, 2025</sup>
### Component name

Enhancement to the existing [`<nys-textinput>` component] (https://its-hcd.github.io/nysds/?path=/docs/components-textinput--docs) component — add optional `end-action` slot

### Component description

Enhance the `nys-textinput` component to support a new optional `end-action` slot. This lets teams embed a contextually relevant button inside the input field. The button would sit flush against the input field border and be aligned to the "inline-end" — using logical CSS properties so that it works as expected with RTL languages. 

### Why the NYS Design System needs this component

Buttons embedded inside a text input is a common pattern across applications and sites, for things like search inputs, address or ZIP lookups, etc. Without this, teams would need to either build new or re-implement this pattern inconsistently or use layout workarounds.

### How the component works

**Slot Name** "action-end" (matches CSS logical property naming)

### Real-world examples

Universal Navigation Bar:
![image](https://github.com/user-attachments/assets/71c90541-cef5-4d07-8c17-7f7ea4d0788c)

### When to use this component and when to consider something else

Using a button inside an input field can improve usability when it provides a clear action **related to the input**.  Ensure the button is intuitive and directly related to the input action.

For example:
Search Fields – A search bar with a magnifying glass icon or a “Search” button inside or at the end of the field. This makes it clear how to submit the query.

### Usability considerations

- **Clear Labels**: Label both input and button.
- **Helpful Hints**: Use placeholders. Do not allow multiple buttons inside the same slot
- **Action Feedback**: Show visual feedback.
- **Keyboard Access**: Ensure tab and Enter/Space work.
- **Button Size**: Buttons should be 44×44px.
- **Error Prevention**: Alert for incorrect inputs.
- **Mobile-Friendly**: Ensure responsiveness.

### Accessibility considerations

1. **Clear and Descriptive Labeling**

- Provide a visible label for the input (do not rely solely on placeholder text).
- Button text must clearly describe its action (e.g., “Search”, “Submit”).
- If no visible label is used, provide an accessible name via aria-label or aria-labelledby.
- Instructions must be clear and present before the user interacts with the field (WCAG 3.3.2).

2. **Keyboard Operability**

- The input and button must be focusable via the Tab key (WCAG 2.1.1).
[...]

PR #513, opened 2025-05-06
NYSDS - [TextInput]: Added the ability to slot in an nys-button

### LINKED ISSUE #500: BUILD: DEVELOP - Textinput Button Slot
### Discussed in https://github.com/ITS-HCD/nysds/discussions/390

<div type='discussions-op-text'>

<sup>Originally posted by **esteinborn** March 20, 2025</sup>
### Component name

Enhancement to the existing [`<nys-textinput>` component](https://its-hcd.github.io/nysds/?path=/docs/components-textinput--docs) component — add optional `end-action` slot

### Component description

Enhance the `nys-textinput` component to support a new optional `end-action` slot. This lets teams embed a contextually relevant button inside the input field. The button would sit flush against the input field border and be aligned to the "inline-end" — using logical CSS properties so that it works as expected with RTL languages. 

### Why the NYS Design System needs this component

Buttons embedded inside a text input is a common pattern across applications and sites, for things like search inputs, address or ZIP lookups, etc. Without this, teams would need to either build new or re-implement this pattern inconsistently or use layout workarounds.

### How the component works

**Slot Name:** "action-end" (matches CSS logical property naming)

### Real-world examples

Universal Navigation Bar:
image

### When to use this component and when to consider something else

Using a button inside an input field can improve usability when it provides a clear action **related to the input**.  Ensure the button is intuitive and directly related to the input action.

For example:
Search Fields – A search bar with a magnifying glass icon or a “Search” button inside or at the end of the field. This makes it clear how to submit the query.

### Usability considerations

- Do not allow multiple buttons inside the same slot

### Accessibility considerations

- The button should be focusable (tabindex="0" if necessary)
- Users should be able to activate it with Enter or Space
- If the button has only an icon (e.g., a search icon), use aria-label</div>

emilygorelik:
# Summary

Allow for a button to be appended to the start or end of the text input.

## Breaking change

`<nys-button>` was modified to have the variables   ` --_nys-button-radius-left:` and `--_nys-button-radius-right` rather than just `--_nys-button-radius`. This change was necessary to be able to override the values independently.

## Related issues

Closes #500 

## Major changes

- **Modified components**: Button radius variables updated
- **Modified components**: Text Input now allows `startButton` and `endButton` slots
- **Documentation updates**: Added documentation for best use on button slot
- 

image
image

### Browser testing

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge/IE

### Accessibility testing

- [x] Keyboard navigation testing
- [x] Color contrast checks


NovaCat35 (dismissed):
LGTM

esteinborn (changes requested):
Can you modify the figma connect file to account for this new button? might need to work with Design to find out how they are implementing so we can modify the code connect file.

emilygorelik:
> Can you modify the figma connect file to account for this new button? might need to work with Design to find out how they are implementing so we can modify the code connect file.

This isn't ready on the design side yet since they cannot suppress props easily

esteinborn:
> > Can you modify the figma connect file to account for this new button? might need to work with Design to find out how they are implementing so we can modify the code connect file.
> 
> This isn't ready on the design side yet since they cannot suppress props easily

Please be sure to run the tests in these two tickets and once complete we can merge:
https://github.com/ITS-HCD/nysds/issues/501
https://github.com/ITS-HCD/nysds/issues/502

Issue #532, opened 2025-05-15
Cannot read properties of undefined (trim()) `_originalErrorMessage`


esteinborn:
When tabbing from out of an input it throws an error message:

*nys-textinput.ts:152
Uncaught TypeError: Cannot read properties of undefined (reading 'trim')
    at _NysTextinput._setValidityMessage (*nys-textinput.ts:152:36)
    at _NysTextinput._validate (*nys-textinput.ts:191:10)
    at _NysTextinput._handleBlur (*nys-textinput.ts:265:10)

![Image](https://github.com/user-attachments/assets/537ef218-1657-46b0-b6cf-48f78e6fefc2)

NovaCat35:
PR https://github.com/ITS-HCD/nysds/pull/538 🎫

PR #538, opened 2025-05-16
Fix: component's trim error

NovaCat35:
# Summary

Fix issue with tabbing out of textinput showing inspector error: **"Cannot read properties of undefined (trim()) _originalErrorMessage"**.

## Breaking change

## Related issues
Ticket https://github.com/ITS-HCD/nysds/issues/532 🎟️


emilygorelik (dismissed):
LGTM, but does this need to be done on other components as well?

NovaCat35:
> LGTM, but does this need to be done on other components as well?

Good call, I went back and checked and made appropriate changes on other components.

esteinborn (approved):
LGTM

PR #542, opened 2025-05-19
New: copy icon

NovaCat35:
# Summary

Add new "copy_content" icon

## Breaking change


esteinborn:
@NovaCat35 please check the list of icons on storybook and the reference website to make sure we have all the icons we should have on the ref site. Thanks

PR #544, opened 2025-05-21
Update button click function to be accessible

### LINKED ISSUE #509: hotfix/button-keyboard-click
Button is not clickable though the keyboard when used as a subcomponent. This issue does not occur when the button is used by itself or is clicked on with the mouse

emilygorelik:
Refactor `<nys-button>` instances used in other components to not need the function called on both `@click` and `@keydown` since we learned about triggering it with `.onClick`

Closes #509


esteinborn (approved):
LGTM

PR #640, opened 2025-06-16
Fix: Button - height overflow when text longer than parent container

emilygorelik:
before:
image

after:
image

the button width is determined by the text inside but if the text is wider than the parent container holding the button the text will wrap, but the height of the button will not accommodate that. By changing the button height to min-height the button can grow in height in these edge cases.


esteinborn (approved):
min-LGTM

PR #670, opened 2025-06-18
fix all clickable issue on checkbox

emilygorelik:
remove wrapping div that caused tile to not me fully clickable


esteinborn (approved):
LGTM

PR #699, opened 2025-07-02
Plop file updates

emilygorelik:
minor changes to make my life easier:

- adds new component to build order
- linting error resolved
- testing uses component instead of HtmlElement which was causing tests to fail initially


NovaCat35 (approved):
LGTM

PR #701, opened 2025-07-03
re-configure imports, cut file size to 380KB

esteinborn:
I do not believe we need to @import all the dependencies between modules. This should be handled by the peerdeps in each components package.json file. 

I have tested this in Storybook, vanilla, and React demos without issue.


emilygorelik (approved):
lgtm

PR #710, opened 2025-07-10
Fix: logic for validateFileHeader

### LINKED ISSUE #709: Bug: Fileinput handleValidation broken
Issue with fileinput broken on accepting MIME types that are not covered. 

If I set mime types to simple ".svg" it works, but if it's in the stack of acceptable mime types it invalidates the file:

 
Image

### LINKED ISSUE #715: Bug: nys-fileinput does not have a main focus ring, and doesnt announce file uploads
NYS-input should have a focus ring around the main container instead of only the upload files button.

It should also read out that it has had a file uploaded to it as well.

Image

Here is an example from USWDS:

Image

NovaCat35:
# Summary
Fixes the issue with the uncovered MIME accepted value, resulting in an error instead of the default passing.
Fixes VO issue.
Fixes the focus state issue.

## Breaking change

## Related issues
Closes #709 🎟️
Closes #715 🎟️

## Major changes
- **Modified components**: `nys-fileinput`

image

### DIFF
--- a/packages/nys-fileinput/src/nys-fileinput.ts
+++ b/packages/nys-fileinput/src/nys-fileinput.ts
@@ -321,9 +321,14 @@ export class NysFileinput extends LitElement {
       const fileItem = this.renderRoot.querySelector(
         "nys-fileitem",
       ) as HTMLElement;
-      if (fileItem) {
-        fileItem.setAttribute("tabindex", "-1");
-        fileItem.focus();
+
+      const innerFileItemMainContainer = fileItem?.shadowRoot?.querySelector(
+        ".file-item",
+      ) as HTMLElement | null;
+
+      if (innerFileItemMainContainer) {
+        innerFileItemMainContainer.setAttribute("tabindex", "-1");
+        innerFileItemMainContainer.focus();
       }
     }
   }


emilygorelik (approved):
lgtm

packages/nys-fileinput/src/validateFileHeader.ts
```diff
@@ -1,159 +1,73 @@
...
-  if (header.length < magic.length) return false;
-  for (let i = 0; i < magic.length; i++) {
-    if (header[i] !== magic[i]) return false;
-  }
-  return true;
-}
-
-// Map of file type aliases
+// Map of accept types to their extensions
 const acceptKeyMap: Record<string, string[]> = {
```
esteinborn, line 2: Are we still checking file types, i thought we were removing that and relying on default "accept" functionality?

esteinborn (changes requested):
We should be removing the accept property and stick with the default functionality of "accept"

NovaCat35:
See here: https://docs.google.com/spreadsheets/d/1IPxA6hkeedWunVbwMbNLCCmWiaEdJ5p5FaTPTLg4xtM/edit?gid=0#gid=0

PR #719, opened 2025-07-22
UnavHeader: fix aria label on search button

emilygorelik:
add aria role to search button on unav header and update nys-button to have prefix and suffix icon as a fallback if necessary


esteinborn (approved):
LGTM, announces the search button properly now.

PR #736, opened 2025-07-23
Fix: icon prop update

### LINKED ISSUE #735: Update `nys-icon` prop
Renaming `label` to `ariaLabel` and removing `ariaDescription` prop on `nys-icon`

NovaCat35:
# Summary

Renaming `label` to `ariaLabel` and removing `ariaDescription` prop on `nys-icon`

## Breaking change

## Related issues

Closes #735 🎟️

### DIFF
--- a/packages/nys-icon/src/nys-icon.stories.ts
+++ b/packages/nys-icon/src/nys-icon.stories.ts
@@ -5,19 +5,18 @@ import "./nys-icon";
 // Define the structure of the args used in the stories
 interface NysIconArgs {
   name: string;
-  label?: string;
+  ariaLabel?: string;
   color?: string;
   rotate?: string;
   flip?: string;
   size?: string;
-  ariaDescription?: string;
 }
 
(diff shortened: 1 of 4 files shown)


packages/nys-icon/src/nys-icon.stories.ts
```diff
@@ -127,18 +123,17 @@ export const NameProp: Story = {
 // Story: Labels
 export const Labels: Story = {
   args: {
-    label: "edit_square icon",
+    ariaLabel: "edit_square icon",
```
emilygorelik, line 126: how does the screen reader parse a underscore? should we have the example label read as "Edit square icon" instead? same comment on file_input example and others as well

emilygorelik (approved):
Approved pending answer about my underscore question

packages/nys-icon/src/nys-icon.stories.ts
esteinborn: Yeah, for production, icon name should have no underscore

PR #751, opened 2025-07-25
Fix: tooltip rendering

NovaCat35:
# Summary
Attempted fix at a problem with the reference site.

Problem with reference site not having focusable for `nys-tooltip`. I suspect this is due to client-side rendering for Storybook and React (which is why it works) versus 11.ty(how HTML is static and render first)

## Breaking change


esteinborn (approved):
LGTM

PR #759, opened 2025-07-29
bugfix/Backtotop appears at top of page

emilygorelik:
Before the BBT was appearing all the time because the scroll check wasn't working, this fixes that as well as corrects the sentence case and button height

demo:
https://github.com/user-attachments/assets/3f960dea-24f2-4969-8ec2-e7aaa8d170bc


emilygorelik:
not fixed yet

esteinborn (approved):
LGTM

PR #770, opened 2025-07-31
Text Input: fix overflowing container bug

emilygorelik:
fix overflow where if width is set to size thats larger than parent container it would overflow


NovaCat35 (approved):
LGTM

PR #783, opened 2025-08-06
Fix: Storybook Dos and Donts Syntax

emilygorelik:
wasnt rendering properly because syntax was wrong
before:
image

after:
image


esteinborn (approved):
LGTM

PR #890, opened 2025-09-16
Enhance: lit analyze checker

emilygorelik:
Add checking script on individual components so we can more easily target fixes


NovaCat35 (approved):
LGTM

PR #894, opened 2025-09-17
NYSDS - RadioButton: fix disabled states

emilygorelik:
before you could click on a radio even if it was disabled, now you cannot :)


esteinborn (dismissed):
LGTM

PR #906, opened 2025-09-18
Enhancement/radio check labels

emilygorelik:
change radio and checkbox to use the `nys-label` under the hood


esteinborn (approved):
LGTM

PR #941, opened 2025-10-03
Hide unloaded components

emilygorelik:
We didn't make sure we added new components as they are built to this check. Now we will


esteinborn (approved):
LGTM

PR #995, opened 2025-10-23
Fix/nysbutton keydown click vanilla

emilygorelik:
Fixed button not being triggered by a keyboard click in vanilla. Update test to catch this if it breaks in the future.


esteinborn (approved):
LGTM;

PR #1024, opened 2025-11-03
Fix: button keyboard

NovaCat35:
# Summary
Make sure the keyboard event for `<nys-button onClick="doFunction();"></nys-button>` works.

WHY does it click but not do keyboard?
```
  /**
   * Vanilla JS & Native HTML keydown solution:
   * The <nys-button onClick="doFunction();"></nys-button> onClick is an attribute here.
   * Thus, we call it here. Otherwise, at this point, this.onClick is null as it isn't prop, but a string attribute
   * In vanilla HTML/JS, clicking executes the attribute function, BUT now with keydown, hence this solution.
   */
```
## Breaking change

### DIFF
--- a/packages/nys-button/src/nys-button.ts
+++ b/packages/nys-button/src/nys-button.ts
@@ -136,9 +136,7 @@ export class NysButton extends LitElement {
 
   private _manageFormAction() {
     // If an onClick function is provided, call it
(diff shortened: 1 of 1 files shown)


esteinborn (approved):
LGTM; testing incoming

PR #1025, opened 2025-11-03
Fix: Unavheader mobile breakpoint "how you know" button

emilygorelik:
Fix the button to be triggered via keyboard but the div to trigger via mouse

https://github.com/user-attachments/assets/3b630b35-359b-497c-a85e-5eb2575709c1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@nysds/components",
-  "version": "1.11.0",
+  "version": "1.11.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@nysds/components",
-      "version": "1.11.0",
+      "version": "1.11.1",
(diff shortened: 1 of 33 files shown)


packages/nys-button/src/nys-button.test.ts
```diff
@@ -68,7 +68,7 @@ describe("nys-button", () => {
 
   it("should trigger click event only once", async () => {
     const el = await fixture<NysButton>(
-      html`<nys-button label="Disabled"></nys-button>`,
+      html`<nys-button label="Button"></nys-button>`,
```
emilygorelik: not critical but fixed to make more sense in the context of the test

esteinborn (dismissed):
LGTM; need to validate it against Robert's button click fix

esteinborn:
Related issue #1027

esteinborn (approved):

PR #1042, opened 2025-11-10
Fix: Step circle was distorted at small screens

emilygorelik:
- also removed a test that was throwing a browser warning


NovaCat35 (approved):
LGTM

Issue #1046, opened 2025-11-12
BUG: `id` attribute not reflected on `nys-` components


NovaCat35:
The `id` attribute is not reflected on `nys-` components. WA allows default reflection; we need to fix this to be reflective

PR #1047, opened 2025-11-12
Fix: reflective `id` property

NovaCat35:
# Summary

Fix `nys-` components to have reflective `id` properties, similar to how Web Awesome allows native default `id` attribute to be reflective. 

Note: I did my best to check the logic and make sure the styling and logic don't break for any `id` that I did remove, as well as adding a postfix `id=${"this.id--label"}` to avoid conflict in `id` attribute duplication.

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1046 🎟️

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@nysds/components",
-  "version": "1.11.1",
+  "version": "1.11.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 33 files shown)


emilygorelik (dismissed):
change the `--label`, otherwise approved

esteinborn (commented):
I need to better understand why we're removing setting the ID on so many elements. Did you run WTR tests on these class name changes? I only see one test file updated

PR #1058, opened 2025-11-17
Fix: Pagination edge case bug

emilygorelik:
Lock totalPages to 1 if an invalid (i.e 0) number is passed in.


esteinborn (approved):
LGTM;

PR #1068, opened 2025-11-24
Fix: Typography token weights

emilygorelik:
`--nys-font-weight-h*` did not exist. All the headings are bols to have them point to the correct token name


esteinborn (approved):
LGTM;

PR #1082, opened 2025-12-03
Fix: properties that take in specific string

emilygorelik:
Refactor props like `width`,`size`, `intent`, etc. that can only take in a specific string to make vscode code completion work better.


esteinborn (approved):
LGTM;

PR #1217, opened 2026-01-28
Enhancement: Description Styling more accessible

emilygorelik:
## after is on the left

image

## changes 
- `scss`ed the style sheets to take advantage of the nesting
- changed `nys-toggle` to use the `nys-label` under the hood and then overrided the font weight, which matches `nys-checkbox` and `nys-radiobutton`

### DIFF
--- a/packages/nys-toggle/src/nys-toggle.ts
+++ b/packages/nys-toggle/src/nys-toggle.ts
@@ -105,6 +105,15 @@ export class NysToggle extends LitElement {
     }
   }
 
+  public formResetCallback() {
+    this.checked = false;
+
+    this._internals.setFormValue(this.checked ? this.value : null);
+
+    // Re-render UI
+    this.requestUpdate();
+  }
+
   /**
    * Event Handlers
    * --------------------------------------------------------------------------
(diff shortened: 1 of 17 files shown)


packages/nys-toggle/src/nys-toggle.ts
```diff
@@ -179,10 +180,16 @@ export class NysToggle extends LitElement {
               </div>
             </span>
           </div>
-          <div class="nys-toggle__text ${this.inverted ? "invert" : ""}">
-            <div class="nys-toggle__label">${this.label}</div>
-            <slot name="description">${this.description}</slot>
-          </div>
+          ${this.label &&
+          html`<nys-label
```
NovaCat35: Even though the `nys-label` here replaces previous code, slot description styling is still being affected by this css line:
```
slot[name="description"] {
    font-style: italic;
    text-align: left;
    display: flex;
}
```

Image

NovaCat35: Additionally, the label is bolder now. Deviates from design

emilygorelik: yes!

PR #1268, opened 2026-02-06
Enhancement/table vars

emilygorelik:
fix/create vars for table to allow for override on the ref site tables


esteinborn (approved):
LGTM;

PR #1288, opened 2026-02-18
Feature: Combobox

emilygorelik:
# Summary
Introducing `<nys-combobox>`

## Breaking change

## Screenshots
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,4 +42,5 @@ Thumbs.db
 
 # Ignore AI assistant files
 CLAUDE.md
+.claude/
 docs/
--- a/package-lock.json
+++ b/package-lock.json
@@ -15286,7 +15286,7 @@
       "devDependencies": {
         "lit": "^3.3.1",
         "typescript": "^5.9.3",
-        "vite": "^7.1.12"
+        "vite": "^7.3.1"
       }
     },
     "packages/nys-datepicker": {
--- a/packages/nys-combobox/package.json
+++ b/packages/nys-combobox/package.json
@@ -26,7 +26,7 @@
   "devDependencies": {
     "lit": "^3.3.1",
     "typescript": "^5.9.3",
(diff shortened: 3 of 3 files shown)


packages/nys-combobox/src/nys-combobox.ts
```diff
@@ -0,0 +1,727 @@
...
+                      ariaLabel="clear selection"
+                      variant="ghost"
+                      size="sm"
+                      circle
+                      @nys-click=${this._handleClearClick}
+                      ?disabled=${this.disabled}
+                    >
+                      <nys-icon
+                        slot="suffix-icon"
+                        size="xl"
```
NovaCat35: Size of the button/icon here seems smaller than what Figma has when compared to the text. I know Figma has icon set to `size='xl'`, but use your discretion to make them align or bigger

packages/nys-combobox/src/nys-combobox.ts
```diff
@@ -0,0 +1,727 @@
...
+      this._internals.ariaRequired = "false";
+      this._internals.setValidity({});
+      this._hasUserInteracted = false;
+    }
+  }
+
+  private _setValidityMessage(message: string = "") {
+    this.showError = !!message;
+
+    if (this._originalErrorMessage?.trim() && message !== "") {
```
NovaCat35: Getting error here when I try to press the chevron down button more than once:

Image

NovaCat35 (changes requested):

esteinborn:
Arrow keys should read the label of the item from the list when you arrow over them
image

esteinborn:
Arrow Keys do not function as expected (They do not allow you to move between items in MacOS VO
Is this an instance where a screen reader user will use tab instead of arrow keys?

esteinborn (approved):
LGTM;

mendeza (approved):

PR #1290, opened 2026-02-18
Fix/stylelint config

emilygorelik:
correct all 3 digit hex codes back to 6

### DIFF
--- a/packages/styles/vite.config.js
+++ b/packages/styles/vite.config.js
@@ -12,6 +12,7 @@ export default defineConfig({
   build: {
     cssCodeSplit: true,
     sourcemap: true,
+    emptyOutDir: false, // Since we're building both ESM and UMD
     rollupOptions: {
       input: {
         // NYSDS Bundles: Using @import to create style bundles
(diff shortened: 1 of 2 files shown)


stylelint.config.mjs
```diff
@@ -3,7 +3,15 @@ export default {
...
-    "selector-id-pattern": null
+    "selector-id-pattern": null,
+    "color-function-alias-notation": null,
+    "color-function-notation": null,
+    "alpha-value-notation": null,
+    "selector-not-notation": null,
+    "property-no-vendor-prefix": null,
+    "declaration-empty-line-before": null,
+    "declaration-block-no-redundant-longhand-properties": null,
+    "color-hex-length": "long",
```
esteinborn: as discussed in our thread, lets standardize the following:

color-function-notation: I think we'd prefer the "modern" setting, and enforce a pattern rather than ignore the value entirely
 
alpha-value-notation: We should standardize on number or percentage, not "null"
 
selector-not-notation: Another standardization, either simple or complex

esteinborn (approved):
LGTM; great work on this!

PR #1408, opened 2026-03-24
Fix: Replace `p` tag with `span` tag

### LINKED ISSUE #1354: BUG: Accordion header has invalid HTML content
Matt O'Brien (@mob411) and Brian Miller (@brianmiller86) have found that the Accordion header fails HTML validity linting performed in their project. 

The below snippet reveals a `<p>` element as child to a `<button>` element.

```html
<button class="nys-accordionitem__heading" type="button" aria-expanded="true" aria-controls="accordionId1-content">
  <p class="nys-accordionitem__heading-title">How do I renew my passport or apply for a new one?</p>
</button>
```

The `<p>` element is [Palpable content](https://html.spec.whatwg.org/dev/dom.html#palpable-content), which is not valid as `<button>` element content. Rather, prefer [Phrasing content](https://html.spec.whatwg.org/dev/dom.html#phrasing-content), for example, a `<span>` element seems to be the most appropriate refactor.

NovaCat35:
# Summary
Replace `p` with `span`.

## Breaking change

## Related issues

Closes #1354 🎟️

### DIFF
--- a/packages/nys-accordion/src/nys-accordionitem.ts
+++ b/packages/nys-accordion/src/nys-accordionitem.ts
@@ -142,7 +142,7 @@ export class NysAccordionItem extends LitElement {
         aria-expanded=${this.expanded ? "true" : "false"}
         aria-controls=${contentId}
       >
-        <p class="nys-accordionitem__heading-title">${this.heading}</p>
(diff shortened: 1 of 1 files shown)


esteinborn (approved):
LGTM;

PR #1427, opened 2026-03-31
Tabs: New Component

### LINKED ISSUE #1315: Code: Tabs

emilygorelik:
image


packages/nys-tab/src/nys-tab.scss
```diff
@@ -0,0 +1,134 @@
...
+
+  .scroll-shadow {
+    position: absolute;
+    top: 50%;
+    transform: translateY(-50%);
+    z-index: 2;
+    opacity: 0;
+    pointer-events: none;
+    transition: opacity 0.2s;
+    height: calc(var(--nys-space-600) + var(--_nys-tab-border-width));
```
NovaCat35: Can you add the fallback for `--nys-space-600`?

packages/nys-tab/src/nys-tab.figma.ts
```diff
@@ -0,0 +1,8 @@
+import figma, { html } from "@figma/code-connect/html";
+
+figma.connect("<FIGMA_TAB>", {
+  props: {
+    // TODO: Add props here
+  },
```
NovaCat35, line 8: Don't forget to add in some code connect props, and connect it on `figma.config.json`. Even if it's still in a Figma branch, at least we know it's working. When the time comes, we just need to change the URL to make it work again.

packages/nys-tab/src/nys-tabpanel.ts
```diff
@@ -0,0 +1,69 @@
...
+  @property({ type: String, reflect: true }) id = "";
+
+  // ---------------------------------------------------------------------------
+  // Lifecycle
+  // ---------------------------------------------------------------------------
+
+  /**
+   * Called when the element is inserted into the document.
+   * Auto-generates a unique `id` if one was not provided, then sets
+   * `role="tabpanel"` on the host so the ARIA role is visible outside the
```
NovaCat35: Looking at your code, there doesn't seem to be any implementation of this logic of setting `role="tabpanel"`

packages/nys-tab/src/nys-tabpanel.ts
```diff
@@ -0,0 +1,69 @@
...
+    this.setAttribute("role", "tabpanel");
+  }
+
+  // ---------------------------------------------------------------------------
+  // Render
+  // ---------------------------------------------------------------------------
+
+  render() {
+    return html`
+      <div class="nys-tabpanel">
```
NovaCat35: Looking over WA code, they also have `aria-hidden="false"` and `aria-hidden="true"` as well as an `aria-labelledby`. Good to implement here

packages/nys-tab/src/nys-tabgroup.ts
```diff
@@ -0,0 +1,391 @@
...
+   * itself.
+   *
+   * Disabled tabs are excluded from navigation and will never receive focus
+   * via arrow keys.
+   *
+   * @param e - The `KeyboardEvent` from the tablist `keydown` listener.
+   * @returns void
+   */
+  private _handleKeydown(e: KeyboardEvent): void {
+    console.log("testing keydown call: ", e.key);
```
NovaCat35: cleanup `console.log()`s

packages/nys-tab/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@nysds/nys-tab",
+  "version": "1.15.1",
```
NovaCat35: Version 1.16.0 :)

packages/nys-tab/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "vite": "^7.3.1"
+  },
+  "keywords": [
+    "new-york-state",
+    "design-system",
+    "web-components",
+    "lit",
+    "nys",
+    "tab",
+    "forms"
```
NovaCat35: Is tab a "forms" component? Remove this keyword if not

packages/nys-tab/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "scripts": {
+    "dev": "tsc --emitDeclarationOnly && vite",
+    "build": "tsc --emitDeclarationOnly && vite build",
+    "test": "vite build && wtr",
+    "build:watch": "tsc --emitDeclarationOnly && vite build --watch",
+    "test:watch": "vite build && wtr --watch",
+    "lit-analyze": "lit-analyzer '*.ts'"
+  },
+  "dependencies": {
+    "@nysds/nys-button": "1.15.1"
```
NovaCat35: version update

packages/nys-tab/src/nys-tabgroup.ts
```diff
@@ -0,0 +1,391 @@
...
+        newIndex = (currentIndex - 1 + tabs.length) % tabs.length;
+        break;
+      case nextKey:
+        newIndex = (currentIndex + 1) % tabs.length;
+        break;
+      default:
+        return;
+    }
+
+    e.preventDefault();
```
NovaCat35: Should `e.preventDefault()` be the first thing called on this function?

NovaCat35 (changes requested):

packages/nys-tab/src/nys-tab.scss
```diff
@@ -0,0 +1,134 @@
...
+  --_nys-button-border-color--disabled: var(--_nys-tab-border-color);
+  --_nys-button-border-color--hover: var(--_nys-tab-border-color--hover);
+  --_nys-button-border-color--active: var(--_nys-tab-border-color--active);
+  --_nys-button-border-radius: var(--_nys-tab-border-radius)
+    var(--_nys-tab-border-radius) 0 0;
+  --_nys-button-border-style: none none solid none;
+  --_nys-button-border-width: var(--_nys-tab-border-width);
+  --_nys-button-padding--x: var(--_nys-tab-padding--x);
+  --_nys-button-padding--y: var(--_nys-tab-padding--y);
+  --_nys-button-line-height: 16px;
```
NovaCat35: Is there a semantic var we can use here?

packages/nys-tab/src/nys-tab.ts
```diff
@@ -0,0 +1,212 @@
...
+   * @returns void
+   */
+  public focus(options?: FocusOptions): void {
+    const button = this.shadowRoot?.querySelector("nys-button") as
+      | (HTMLElement & { focus?: (o?: FocusOptions) => void })
+      | null;
+    if (button?.focus) {
+      button.focus(options);
+    } else {
+      super.focus(options);
```
NovaCat35: Is this necessary? I feel like the `nys-button` is always in shadowDOM.

NovaCat35 (changes requested):

packages/nys-tab/src/nys-tabpanel.ts
emilygorelik: It is set in the connected callback on line 51

mendeza:
@emilygorelik this is a very nice start. Love the `Arrow` key wraparound behavior in the tabheaders.
Only two changes to make.
1. The headers have a focus trap. You can `Tab` into them, but then can't `Tab` out in either direction. Trapped!
2. @schugartl and I concur that the content panels should have `tabindex="0"`

Combining these two, a user could:
1. `Tab` from page above &rarr; selected tabheader &rarr; tabcontent boundary &rarr; first focusable element in tabcontent &rarr; down the page
2. And the exact reverse: `Shift-Tab` from below the tabset &rarr; focusable element in tabcontent &rarr; tabcontent boundary &rarr; selected tabheader &rarr; next focusable element in page content

mendeza:
Hi @emilygorelik please adjust the header tabindices like so: currently selected has either no tabindex attr (since headers are buttons, and buttons are focusable by default, they don't need it) or `tabindex="0"`. Non-selected headers are `tabindex="-1"`. This will solve the problem of tab order.

tabset with headers annotated with appropriate tabindex values

This is as good a time as any to fix the problem at its source!

screen of button demo next to devtools inspector showing the attribute tabindex='0' is set by default on the NYS button

esteinborn:
@emilygorelik my review:

 - [ ] MacOS VO: user tabs to first tab, lands on the tab, tab says "1 of 6," but you never get to the 6th tab because it is disabled. I believe we said that disabled controls should still be focusable, but identify that they are in fact disabled.

 - [ ] Chrome: tabbing into the tab list locks the user focus into the tab list

 - [ ] Chrome: no way to horizontally scroll with a mouse at smaller resolutions. Trackpad works with a two finger swipe left and right, but mouse users have no recourse. DO NOT scroll-jack the mouse here, consider a click and drag approach(?) open to any ideas to resolve this.

mendeza:
> * but you never get to the 6th tab because it is disabled. 

Hi @esteinborn are you using `Arrow` navigation? NVDA/Win (via `Arrow`) does arrive at the disabled header and indicates it's unavailable

mendeza:
> * DO NOT scroll-jack the mouse here, consider a click and drag approach(?)

Hi @esteinborn Presumably mouse users can use `Arrow` keys. Out of curiosity, what's your objection to the scroll wheel?

esteinborn:
> > * but you never get to the 6th tab because it is disabled.
> 
> Hi @esteinborn are you using `Arrow` navigation? NVDA/Win (via `Arrow`) does arrive at the disabled header and indicates it's unavailable

Yes, using Macos VO with arrow keys and it does not hit the disabled tab to read it out or indicate in any way that it is disabled


> > * DO NOT scroll-jack the mouse here, consider a click and drag approach(?)
> 
> Hi @esteinborn Presumably mouse users can use `Arrow` keys. Out of curiosity, what's your objection to the scroll wheel?

NNG has a great breakdown of scroll-jacking and how disorienting it can be to users. Not to mention that there is only the slight "raised" gradient at the right or left indicating other tabs, but is not fully clear about it. https://www.nngroup.com/articles/scrolljacking-101/

mendeza:
> Yes, using Macos VO with arrow keys and it does not hit the disabled tab to read it out or indicate in any way that it is disabled

@esteinborn, yes I confirm your experience. In terms of baseline cross-SR experience, VO definitely has a poorer experience than NVDA with the current state of the implementation. For another example, when I activate a header, NVDA immediately reports "Selected," VO does not. However it will somewhat belatedly report this when one navigates to a header that is already selected.

Regarding the NNG scrolljacking piece, was it [Scrolljacking 101](https://www.nngroup.com/articles/scrolljacking-101/)?
I completely agree with you that its points on full-page scrolljacking are irrefutable. But I also think the article points out how using the technique in a very limited context may be OK and useful. Anecdotally, while interacting with the component with a mouse, I tried the scrollwheel practically without thinking, because I _expected_ that to work.

From the article:

>**Scrolljacking Worked Best When It Had Functional Value**
Scrolljacks were best tolerated when they were used to progressively disclose relevant information. For example, on the BBC News website most study participants tolerated the scrolljack well and even benefited from it, because it added contextual information. However, it still disoriented some participants, causing them to hesitate and double check their place on the page.
>
>**The Scrolljack Should Progressively Disclose Valuable Information**
 A scrolljack is beneficial when it gradually adds supporting information to lessen cognitive load and support storytelling. When it does not serve this function, it is not worth it: it requires more work for the user with no benefit.
>
>Ensure that scrolljacking is used only for showcasing critical pieces of information. For example, on the Apple website, scrolljacking is used to break down the features of the Apple Watch Ultra and to map them to the physical device. To determine whether the scrolljack has a functional purpose in your design, ask the following questions:
>
>- What purpose does scrolljacking serve?
>- Does it add relevant context?
>- Does it break down a complex concept?
>- Does it communicate something about your brand?
>- To justify implementation costs, do we have research to support that this functional purpose is valuable to users?
>- What meaning is lost when the scrolljack is removed?

HOWEVER, it may be a moot point: as this video shows, on desktop `Shift` becomes a modifier to the scroll wheel, allowing scrolling of content overflowing horizontally. And as noted previously, the `Arrow` keys work.

IF we want to get more hardcore, there is the classic "collapse into 'More'" pattern of a dynamic dropdown. 
We can see both of these approaches in the GitHub UI:

https://github.com/user-attachments/assets/9c2c9c71-dea7-4e4b-be9c-3f5e5689674b

esteinborn:
Branch is failing build.

mendeza:
A11y-wise, we consider this good-to-go.

mendeza:
I've just been doing some experimentation with `name`, `aria-label` and `aria-labelledby` since unfortunately there is poor SR-UX baseline between macOS VoiceOver and Win NVDA in terms of the `name` value being announced. This is not any reason to hold back merging v1 of the component, but experimentation should continue as part of ElementInternals. For example, @NovaCat35, we could try a value convention pattern that could be used widely if it works:

```
<h3 id="claudine">Claudine is not Claude</h3>

<!-- 1 -->
<nys-component name="Claudine">
<!-- renders...-->
<div aria-label="Claudine">
<!-- SR announces: "Claudine" -->
<!-- 2 -->
<nys-component name="#claudine">
<!-- renders...-->
<div aria-labelledby="claudine">
<!-- SR announces: "Claudine is not Claude" -->
```

mendeza:
corrected above
```
<nys-component name="#claudine">
<!-- renders...-->
<div aria-labelledby="claudine">
```

mendeza:
Since the `name` support is wonky with SRs (even the APG tablist examples struggle with this), the site demos would benefit from good content structure examples, such as a heading element.

packages/nys-tab/src/nys-tabgroup.ts
```diff
@@ -0,0 +1,395 @@
...
+ * {@link https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ ARIA Tabs Pattern}:
+ * - Arrow keys move focus without changing selection.
+ * - Enter / Space confirm selection on the focused tab.
+ *
+ * @element nys-tabgroup
+ *
+ * @slot - Accepts `<nys-tab>` and `<nys-tabpanel>` children. Elements are
+ *   moved into internal shadow-DOM containers on `slotchange`; the slot
+ *   itself is not rendered visibly.
+ */
```
esteinborn, line 54: Should add an `@example` of a tab group properly laid out for the MCP/CEM

see: [Stepper example](https://github.com/ITS-HCD/nysds/blob/f55323a83f5c3c49cbe6512939dd5727260d57cc/packages/nys-stepper/src/nys-stepper.ts#L22)

esteinborn (changes requested):

PR #1442, opened 2026-04-02
Button: CSS inverted border missing

### LINKED ISSUE #1418: Bug Button Border color var not properly set
Image

emilygorelik:
border color hover and active states were missing a default value and were rendering incorrectly in inverted filled variant 

closes #1418


esteinborn (approved):
lgtm

PR #1448, opened 2026-04-07
Button: simplify border-radius prop

emilygorelik:
simplify internal `--_nys-button-border-radius` variable. Not a breaking change since only used internally.


NovaCat35 (approved):
LGMT

esteinborn (approved):
LGTM;

PR #1471, opened 2026-04-22
Fix: bundle CEM and token data into mcp-server package

plasticmind:
## Summary

- Adds a `copy-data` build script that copies `custom-elements.json` and token files into `packages/mcp-server/data/` at build time
- Updates `cem-parser.ts` and `token-parser.ts` to check the bundled data path first, before falling back to monorepo-relative paths
- Adds the generated `data/` directories to `.gitignore`

## Why

Users who installed `@nysds/mcp-server` standalone (without the full NYSDS monorepo installed globally) were getting empty responses because the CEM and token JSON files weren't included in the published package. This fix bundles those files directly into the package so it works in any install context.

## Test plan

- [ ] Install `@nysds/mcp-server@next` in a fresh environment (no NYSDS globally installed)
- [ ] Confirm MCP tools return component and token data
- [ ] Confirm monorepo-local usage still works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,10 @@ Thumbs.db
 # Turborepo cache
 .turbo
 
+# MCP server bundled data (copied at build time)
+packages/mcp-server/data/cem/
+packages/mcp-server/data/tokens/
+
 # Ignore AI assistant files
 CLAUDE.md
 .mcp.json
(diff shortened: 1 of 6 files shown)


esteinborn (approved):
LGTM;

Issue #1475, opened 2026-04-23
BUG: `nys-datepicker` `required` error message not align with other form components


NovaCat35:
## Summary
The `nys-datepicker` property, when set to `required` error message does not align with other form components.

## Screenshot
<img width="631" height="303" alt="Image" src="https://github.com/user-attachments/assets/ba159f1c-5a4b-49b9-9a5a-4f3d0b4d6af8" />

PR #1476, opened 2026-04-23
Fix: align `nys-datepicker` error message with other form related components

NovaCat35:
# Summary

Fix bug where `nys-datepicker` required error message does not align with other form components.

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1475 🎟️


esteinborn (approved):
LGTM

PR #1478, opened 2026-04-27
Fix: utility classes

### LINKED ISSUE #1088: [Figma] Update to Properties on Global Header
The "Content" dropdown has been changed to "Title Options" after receiving feedback that it wasn't clear that you can change the layout of the application and/or agency name. Arrow icons were also added to the associated text fields. This will need to be updated in order for code connect to work. 

Image

### LINKED ISSUE #1382: Bug: Responsive utility classes fail due to CSS source order / specificity conflict
## Bug Report

Responsive utility class variants (e.g. `nys-desktop:nys-display-block`) can be overridden by base utility classes (e.g. `nys-display-none`) due to identical CSS specificity and unfavorable source order in the compiled stylesheet.

## Root Cause

The `responsive-variants` mixin in `packages/styles/src/utilities/_mixins.scss` generates all variants (base + all breakpoints) for each value before moving to the next value. This means the compiled CSS is ordered **by value**, not **by breakpoint**:

```css
/* Generated order (simplified): */
.nys-display-block { display: block; }
@media (min-width: 64em) { .nys-desktop\:nys-display-block { display: block; } }

.nys-display-none { display: none; }
@media (min-width: 64em) { .nys-desktop\:nys-display-none { display: none; } }
```

Since both `.nys-display-none` and `.nys-desktop\:nys-display-block` have specificity **(0,1,0)** — media queries do not affect specificity — the rule appearing **later in source order wins**. Because `none` is generated after `block` in the `$display-values` list, `display: none` always beats `display: block` regardless of viewport width.

## Affected Utilities

All utilities using the `responsive-variants` mixin are affected:
- **display** (`_display.scss`)
- **opacity** (`_opacity.scss`)
- **z-index** (`_zindex.scss`)
- **flex** (`_flex.scss`)
- **float** (`_float.scss`)
- **order** (`_order.scss`)
- **overflow** (`_overflow.scss`)
- **position** (`_position.scss`)

## Reproduction

```html
<!-- FAIL: none is declared after block in source, so none always wins -->
<div class="nys-display-none nys-desktop:nys-display-block">
  Should be visible at 1024px+ but is always hidden
</div>

<!-- FAIL: opacity-100 is declared after opacity-30, so 100 always wins -->
<div class="nys-opacity-100 nys-desktop:nys-opacity-30">
  Should be translucent at 1024px+ but is always fully opaque
</div>

<!-- OK: inline-block is declared after none, so this works by accident -->
<div class="nys-mobile-lg:nys-display-none nys-desktop:nys-display-inline-block">
  Works only because inline-block comes after none in source order
</div>
```

## Recommended Fix

Restructure the `responsive-variants` mixin (or the way utilities consume it) so that the compiled CSS is grouped **by breakpoint** rather than **by value**. All base classes should be emitted first, followed by all `mobile-lg` variants, then all `tablet` variants, etc.:

```css
/* Desired order: */
.nys-display-block { display: block; }
.nys-display-none { display: none; }

@media (min-width: 30em) {
  .nys-mobile-lg\:nys-display-block { display: block; }
  .nys-mobile-lg\:nys-display-none { display: none; }
}

@media (min-width: 64em) {
  .nys-desktop\:nys-display-block { display: block; }
  .nys-desktop\:nys-display-none { display: none; }
}
```

This ensures that at any given breakpoint, the responsive variant always appears later in source order than the base class, so it correctly overrides. This is the standard pattern used by Tailwind CSS and other utility-first frameworks.

The change would be in `packages/styles/src/utilities/_mixins.scss` and potentially in how each utility file calls the mixin.

NovaCat35:
# Summary

Fix issue with responsive utility class variants (e.g. nys-desktop:nys-display-block) can be overridden by base utility classes (e.g. nys-display-none) due to identical CSS specificity and unfavorable source order in the compiled stylesheet.

## Breaking change
Please double check

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1088 🎟️
Closes https://github.com/ITS-HCD/nysds/issues/1382 🎟️


packages/styles/src/utilities/_display.scss
```diff
@@ -7,7 +7,17 @@
 @use "mixins" as *;
 
 @each $value in $display-values {
-  @include responsive-variants("nys-display-#{$value}") {
+  @include base-class("nys-display-#{$value}") {
```
esteinborn, line 10: what is the enew `base-class` meant to represent?

esteinborn (changes requested):
Please explain `base-class` so I understand the change.

packages/styles/src/utilities/_display.scss
NovaCat35: This is to separate the responsive-variants() CSS, where it was previously doing both the assigning of the base utility classes values and the responsive queries.

This separation of concern fixes the bug

esteinborn (changes requested):
@NovaCat35 Please rebuild the styles package and fix the sass errors that appear. Then we can look at merging this.

PR #1486, opened 2026-04-30
Hot Fix: Add space keyboard event for `nys-breadcrumbs` ellipses

NovaCat35:
# Summary
Add an event listener for "space" to account for space expanding out the `nys-breadcrumbs` ellipses collapse state

## Breaking change


esteinborn (approved):
LGTM;

Issue #1498, opened 2026-05-04
BUG: `nys-tooltip` when programmatically changing text goes back to top-left corner of page


NovaCat35:
## Summary
`nys-tooltip` when programmatically changing text goes back to top-left corner of page

PR #1505, opened 2026-05-06
Fix: `nys-tooltip` issue with text change resetting tooltip position

NovaCat35:
# Summary
Fix: Programmatically changing text goes back to the top-left corner of the page

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1498 🎟️


esteinborn (changes requested):
Please do a checkout of custom-elements.json from the develop branch and merge that in. There shouldn't be any tab cem regen happening outside of that PR @NovaCat35

Issue #1002, opened 2025-10-27
Enhancement: `nys-alert` no heading format


NovaCat35:
## Summary
Emily: "slotting in text with no heading is formatted oddly"
Robert: "Heading is required, I should doc that" ... "Actually hmmm, I can see what I can do to make this a reality if we don't want header..."

<img width="1015" height="120" alt="Image" src="https://github.com/user-attachments/assets/416d5acf-f93c-429a-8b9d-f3c2e5eb35b0" />

PR #1507, opened 2026-05-07
Fix: Remove heading section when no prop provided

NovaCat35:
# Summary
Remove heading from rendering when no `heading` prop for `nys-alert`

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1002 🎟️


esteinborn (approved):
LGTM;

PR #1511, opened 2026-05-08
Fix: Grid gap overlaps running off screen

### LINKED ISSUE #1499: Vertical spacing for grids, similar to grid-gap which is currently only the left/right padding

### LINKED ISSUE #1509: BUG: grid-container and grid-gap need to be unified
Currently, on some pages on the reference website, there is a small amount of h-scroll on the page.

Image

This is caused by the following:

# Root Cause

`nys-grid-gap-6` uses -1.5rem (24px) negative margins on the grid row, but `nys-grid-container-widescreen` only has 1rem (16px) padding at viewports below 1024px.

This causes the row to extend 24px - 16px = 8px beyond the viewport on the right side, producing a visible horizontal scrollbar.

## Fix

Unify the margins so they do not add more negative margin than was set on the container class.

emilygorelik:
fix overflowing large grid gap classes on small screens by forcing them to use a smaller gap. 
Also closes #1499 by adding in `nys-grid-gap-x-ramp` and `nys-grid-gap-y-ramp`. The default `nys-grid-gap-ramp`now applied the gap to horizontal and vertical padding.
image
image


packages/styles/src/utilities/_layout-grid.scss
```diff
@@ -25,6 +25,7 @@
     max-width: $max-width;
     padding-left: 1rem;
     padding-right: 1rem;
+    overflow-x: clip;
```
esteinborn: lets not overflow clip. lets look at all the margins and negative margins for the grid and unify them.

esteinborn (changes requested):
overflow clip this is not.

packages/styles/src/utilities/_layout-grid.scss
```diff
@@ -77,58 +71,222 @@
...
+// Maintains flexbox layout math: negative margin on row + padding on children
+// prevents gaps from consuming space, preserving column width calculations.
+// Native gap would cause column wrapping (e.g., 3×col-4 + gap > 100%)
 // ============================================
 
 // Default gap (responsive: 0.5rem base, 1rem at 64em+)
 .nys-grid-row.nys-grid-gap {
-  margin-left: -0.5rem;
-  margin-right: -0.5rem;
+  margin: -0.5rem;
```
esteinborn: went from margin left-right only to all margin? is this intended?

packages/styles/src/utilities/_config.scss
```diff
@@ -30,26 +30,6 @@ $grid-containers: (
...
-  "-250": 1.25rem,
-  "-300": 1.5rem,
-  "-400": 2rem,
-  "-500": 2.5rem,
-  "-600": 3rem,
-  "-700": 3.5rem,
-  "-800": 4rem,
-  "-1200": 6rem
-);
-
```
esteinborn, line 52: Are we supposed to be removing all the grid-gaps?

esteinborn (changes requested):
Please address the comment re: grid-gaps being removed from config.scss file

esteinborn (dismissed):
LGTM;

PR #1512, opened 2026-05-11
A11y/back to top

### LINKED ISSUE #1467: [A11y] Back To Top improvement request
# Overview

* Back To Top benefits only sighted mouse-users. It is non-operational for all other user personas. Specifically, actual focus is not moved, which makes this a focus order violation of [WCAG SC 2.4.3](https://www.w3.org/WAI/WCAG22/quickref/?showtechniques=243#focus-order).
  * **Request**: programmatically move focus to `document.body`. Why `body`? We cannot assume what every user will consider an ideal "top." For example, return to `main` may frustrate users seeking content above it. Return to `body` places the keyboard or SR user a single keystroke away from the page's Skip Link.

emilygorelik:
Enhance docs to clearly outline proper placement. Already a11y tested by @mendeza. Will require an audit of ref site to ensure new page layout structure is followed so screenreaders can properly focus the BTT button

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -1058,7 +1058,7 @@
       "declarations": [
         {
           "kind": "class",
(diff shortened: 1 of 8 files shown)


esteinborn (approved):
LGTM;

PR #1534, opened 2026-05-15
Fix: a11y issue for avatar

### LINKED ISSUE #1466: [A11y] Avatar accname improvement and interactive fix requests
# Overview 

* the `ariaLabel` prop is too liberal in what it accepts:
  * `" "` empty space and `&nbsp;` are passed through. This is not robust.
  * When `aria-label` is omitted entirely, a default value of "avatar" is rendered. This is useless, like announcing, "this is a button button."
  * **Request**: make this prop _required_ and disallow suspect values.
 * With [`interactive`](https://designsystem.ny.gov/components/avatar/#interactive) set, Avatar becomes partially a fake button: it has `role=button tabindex=0`. When an `onClick()` function is attached, an interactive Avatar responds **only** to mouse click. However it **must** also respond to <kbd>Enter</kbd> and <kbd>Space</kbd>. 
   * **Request**: fake buttons are technical debt waiting to bite you. Rather than continue down that path, _use the_ [`nys-button circle`](https://designsystem.ny.gov/components/button/#circle) _under the hood_. It would need little other massaging aside from size reduction.

NovaCat35:
# Summary

Fix a11y issue for `nys-avatar` listed in the [related ticket](https://github.com/ITS-HCD/nysds/issues/1466).

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1466 🎟️


esteinborn (approved):
LGTM;

Issue #1463, opened 2026-04-16
Enhancement: Remove performance heavy resize event in JS


NovaCat35:
## Summary
If you filter for `window.addEventListener("resize"` on the NYSDS repo, you'll find several components rely on this functionality.

Please first test the expected responsive results and account for an alternative solution that is less performance taxing. Implement with CSS + media queries instead of JS to prevent heavy performance issues from constantly polling resize event.

See article: 
- https://blog.openreplay.com/avoiding-resize-event-pitfalls-js/
- https://tigeroakes.com/posts/resize-event-alternatives/

mendeza:
You can throttle, but then may end up tweaking individual instances to tastes.

This is a nice write-up: https://tigeroakes.com/posts/resize-event-alternatives/; among others it references:
- [Resize Observer](https://developer.mozilla.org/en-US/docs/Web/API/Resize_Observer_API) (the modern way, may be the easiest refactor, definitely A-B the perf)
- [matchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) (if media queries are sufficient; I've been wanting to try this one)

PR #1541, opened 2026-05-18
Enhancement: remove performance heavy resize listeners

NovaCat35:
# Summary

The previous codebase used a number of `addEventListener("resize", ...)` calls, which can be performance-taxing.

These have been removed and replaced with `matchMedia` and `ResizeObserver.`

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1463 🎟️

## Testing
Tested on Storybook

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -2779,6 +2779,11 @@
             },
             {
               "kind": "field",
+              "name": "_mobileQuery",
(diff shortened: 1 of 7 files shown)


esteinborn (approved):
LGTM;

PR #1549, opened 2026-05-20
Enhancement/stepper docs

emilygorelik:
Update js docs to help LLMs implement better. Also one small fix with how the selected and current state is being updated to work with react better. Tested in both vanilla and react.


esteinborn (approved):
LGTM;

PR #1568, opened 2026-05-26
Bug/textinput type tel reset

### LINKED ISSUE #1567: Bug: textinput type tel reset
Hi, I'm trying to use an NYSDS button of type reset. I have a nys-textinput field of type "tel" on my screen. It starts off blank with some watermark type characters

When I hit reset, it does clear what I have in there but then it sets the data I had in there as a new watermark text

emilygorelik:
Fix bug causing mask to not reset properly when resetting a form


esteinborn (approved):
LGTM;

PR #1597, opened 2026-06-04
Enhancement/vertical grid gap

### LINKED ISSUE #1499: Vertical spacing for grids, similar to grid-gap which is currently only the left/right padding

emilygorelik:
Make grid gap padding apply to both x and y by default.
introduce `grid-gap-x` and `grid-gap-y` classes

> [!WARNING]  
> This is a breaking change since now the `grid-gap` classes will apply a vert padding as well. will need to convert them to `grid-gap-x` if want to remain only horizontal

image
image
image


esteinborn (approved):
Approved, please add a line to the sprint 44 ticket for this release for this doucmentation

PR #1611, opened 2026-06-12
Remediate v1.18.1 audit findings (consolidated, rebased on 1.19.0)

plasticmind:
Ran a full audit of 1.18.1 across security, accessibility, and tooling — 89 findings total. Here's what changed and why.

**`nys-button` could execute arbitrary code.** It took an inline `onclick="..."` attribute and ran it through `new Function()`, effectively `eval()`. This broke the component under strict Content Security Policies and created an injection vector. The fix dispatches a native click and lets the browser run the handler normally. Added a `no-new-func` lint rule to prevent regressions.

**`nys-backtotop` broke server-side rendering.** It called `window.matchMedia()` in its constructor, which throws in any SSR context (Next, Astro, Node) before the page renders. Guarding with `typeof window` fixes this, and a new `test:ssr` check catches it at CI.

**Missing `sideEffects` field risked components being tree-shaken away.** Without it, an aggressive bundler can strip the custom-element registration, since nothing appears to import it. Every package now declares its side effects explicitly.

**Three components had accessibility gaps.** Breadcrumbs missing `aria-current`, a datepicker error not associated with its input, and a menu without an accessible name. All fixed and tested.

The rest is tooling: `npm ci` for reproducible installs, a security-audit gate, Dependabot, SHA-pinned actions, and a `lit-analyze` glob that was only checking one file.

**Not included:** anything 1.19.0 already fixed (rebased onto it and dropped duplicates), the Figma token rotation (an ops task, not a code change), and the P2 long tail — form-callback gaps, remaining WCAG issues, and bundle-size work. Those warrant their own PRs.

No breaking changes. `build:all` passes and `npm audit` reports 0 vulnerabilities. Leaving versioning to the maintainers' release process.

---

Consolidates the v1.18.1 security / a11y / tooling audit remediation onto current `develop` (1.19.0). Originally executed across parallel git worktrees; rebuilt here as a single branch — dropping work that 1.19.0 already absorbed and completing the side-effects workstream that was never executed.

## Workstreams

- **Security (P0):** Remove the `new Function()` code-injection vector in `nys-button` (CVSS 8.1); inline `onclick` now delegates to a native `this.click()`. Adds a `no-new-func` ESLint guard to prevent recurrence. *(The vector was still present in 1.19.0.)*
- **WCAG a11y:** ARIA correctness fixes + tests for `nys-breadcrumbs`, `nys-datepicker`, and `nys-dropdownmenu`.
- **SSR:** `typeof window` / `typeof document` guards in `nys-backtotop` (moves `matchMedia` out of the constructor) so the bundle is safely importable server-side. Adds a `test:ssr` Node-import check.
- **CI:** `npm ci`, `npm audit --audit-level=high`, and a Dependabot config. *(SHA-pinning of actions landed independently in 1.19.0; the deploy workflow is now pinned too.)*
- **Tooling:** Plop now keeps `tsconfig.build.json` in sync on `npm run gen`; `lit-analyze` glob fixed (`'*.ts'` → `'src/**/*.ts'`); package template emits `sideEffects`.
- **Tabgroup:** Adds disconnect-cleanup tests for `nys-tab` (the underlying fix shipped in 1.19.0).
- **Tree-shaking:** Adds a `sideEffects` field to every package so bundlers preserve custom-element registration (`true` for the 33 component packages + root; `false` for tokens/mcp-server; `["**/*.css"]` for styles).
- **Test DX:** Decouples Playwright install from the test hot path via a `pretest` hook (cached local runs ~1s instead of `--with-deps` on every run).

## Notes

- **No breaking changes.** All changes are bug fixes, additive ARIA attributes, SSR-safe guards, or build metadata.
- `npm run build:all` passes; `npm audit` reports **0 vulnerabilities**.
- Versioning is left to the maintainers' release process (1.19.0 shipped via a dedicated Release PR).


.github/workflows/deploy-github-pages.yaml
esteinborn, line 1: This file doesn't need an update. the current changes revert an upgrade to v6 of checkout and setup-node

.github/workflows/test-node-js.yaml
```diff
@@ -18,13 +21,15 @@ jobs:
...
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v6
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
       with:
         node-version: ${{ matrix.node-version }}
     - name: Install dependencies
-      run: npm i
+      run: npm ci
+    - name: Security audit
+      run: npm audit --audit-level=high
```
esteinborn: I'd rather have this as a locally run check. I have run into multiple instances where we have a vulnerability in a nodeJS package, and `npm audit --fix` doesn't fix the root cause, or its a major version upgrade. 

I agree we should be auditing more than just seeing a list of X vulnerabilities when we `npm install`

.github/workflows/test-node-js.yaml
```diff
@@ -18,13 +21,15 @@ jobs:
         node-version: ['24.15.0']
 
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v6
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
```
esteinborn: lets revert these back to v6

packages/nys-alert/package.json
```diff
@@ -40,5 +40,6 @@
     "alert"
   ],
   "author": "New York State Design System Team",
-  "license": "MIT"
+  "license": "MIT",
+  "sideEffects": true
```
esteinborn, line 44: Shouldn't the individual modules use sideEffects: false, since we do not need to bundle every JS function for every component.

we shoudl be aware of [Common pitfalls with sideEffects](https://webpack.js.org/guides/tree-shaking/#common-pitfalls-with-side-effects)

src/templates/package.template.hbs
```diff
@@ -11,6 +11,7 @@
     }
   },
   "type": "module",
+  "sideEffects": true,
```
esteinborn, line 14: See previous comment RE: sideEffects. I think we want to set these as "false" which is what triggers the tree-shaking for the individual component

esteinborn (changes requested):
Couple of things to discuss, but mostly looks good.

esteinborn (approved):
LGTM;

Issue #1616, opened 2026-06-16
Enhancement: `@layer` cascade specificity hierarchy


NovaCat35:
## Summary
Add `@layer` cascade hierarchy to global stylesheets so that component-level styles can reliably override base styles without specificity battles or relying on source order.

esteinborn:
Lets regroup on this @layer thing. It may not be what we need to solve our current problem of trying to open up the shadowDOM thereby exposing styling to outside CSS

PR #1617, opened 2026-06-16
Enhancement: Add `@layer` cascade hierarchy to global stylesheets

NovaCat35:
# Summary
Add `@layer` cascade hierarchy to global stylesheets so that component-level styles can reliably override base styles without specificity battles or relying on source order.

## Breaking change
Marked as potentially breaking because adding cascade layers changes how styles override each other. App developers may now be able to override reset and base styles by controlling layer order instead of relying on specificity or file order.

**Initial testing did not show issues, but it needs a closer look for edge cases in existing apps.**

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1616 🎟️

### DIFF
--- a/packages/styles/src/nysds.scss
+++ b/packages/styles/src/nysds.scss
@@ -10,17 +10,17 @@
    * Version: 1.19.0
 */
 
+/* Include Design Tokens (variables for colors, spacing, typography) */
+@use "./core/tokens";
+
 /* 
  * Cascade layer order — later layers have higher priority.
  * Unlayered consumer CSS always wins over all layers. 
  */
-@layer nysds.base, nysds.components, nysds.utilities;
-
-/* Include Design Tokens (variables for colors, spacing, typography) */
-@use "./core/tokens";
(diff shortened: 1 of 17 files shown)


packages/styles/src/nysds.scss
```diff
@@ -10,9 +10,18 @@
...
+ * Cascade layer order — later layers have higher priority.
+ * Unlayered consumer CSS always wins over all layers. 
+ */
+@layer nysds.base, nysds.components, nysds.utilities;
+
 /* Include Design Tokens (variables for colors, spacing, typography) */
 @use "./core/tokens";
 
+/* Include NYSDS component styles (Light DOM) */
+@use "./components";
```
NovaCat35: This `@use "./components";` is for the upcoming stylesheets of `nys-` components that will have `.scss` files targeting light DOM slots (for example, see vertical navigation code/PR)

esteinborn:
Closing as "won't do"

Issue #1614, opened 2026-06-15
Bug: Datepicker default width is medium, should be full


emilygorelik:
This is a stylistic breaking change but we need date picker to match other form input components. Be sure to document this as a breaking change in the release notes

PR #1625, opened 2026-06-17
Fix:  set `nys-datepicker` default width to `full`

NovaCat35:
# Summary

Change `nys-datepicker` default width from `md` to `full`

## Breaking change

⚠️ This is a **breaking** change.  

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1614 🎟️


esteinborn (approved):
LGTM;

Issue #1626, opened 2026-06-17
BUG: Drupal support for `nys-checkbox`


NovaCat35:
## Summary 
Drupal uses bracket notation on our NYSDS components. In this case, with `nys-checkbox` it uses

``` html
<nys-checkboxgroup label="Pick colors" required>
  <nys-checkbox name="colors[red]"   value="Red"   label="Red"></nys-checkbox>
  <nys-checkbox name="colors[blue]"  value="Blue"  label="Blue"></nys-checkbox>
  <nys-checkbox name="colors[green]" value="Green" label="Green"></nys-checkbox>
</nys-checkboxgroup>
```

Issue is this approach has issue where `nys-change` event is dispatched to group which is calling a function `_handleCheckboxChange()`. This function is overriding the fieldset's names improperly. Fix it.

[Troubleshooting Thread Link](https://teams.microsoft.com/l/message/19:0228156e2bb5419c8152047f596a7bfb@thread.tacv2/1781019859866?tenantId=f46cb8ea-7900-4d10-8ceb-80e8c1c81ee7&groupId=40dc9e8f-13b9-4301-8cb7-db5c2b21c9fa&parentMessageId=1781019859866&teamName=NYS%20Design%20System&channelName=Troubleshooting&createdTime=1781019859866)
> Title: checkbox and checkboxgroup events
> Developer: Fasoldt, Adam Vincent (ITS)
> 
> I am having some issues with checkboxgroups. In our system (Drupal), the name props of each checkbox in > a checkbox group use the checkboxGroupName[checkBoxName] format for their individual checkbox name properties. The NYSDS seems to (fairly correctly) be expecting the names of all checkboxes in a group to be the same. It is valid to use [brackets] in this way for various kinds of processing. Whenever a checkbox is checked, the nys-changed event fires and some code in the checkbox js changes the name of the group's fieldset to that checkbox's name. So, our <fieldset> names end up being like colors[blue] instead of colors. When I block nys-changed from firing, this no longer occurs.
>  
> Even after blocking nys-changed, the value of the fieldset is set and I'm not really sure where it's coming from and I'm not 100% sure it's from the design system, but I'm fairly sure it is. I saw a setFormValue function, but I didn't delve too deep into what it is. However, the problem is that our system sends an array of values to the backend for processing while the value set by (I think?) the design system code is a comma-delimited string. 
>  
[...]

PR #1631, opened 2026-06-23
Fix: Drupal `name` fix on `nys-checkboxgroup`

NovaCat35:
# Summary
Fixed `nys-checkboxgroup` overwriting the fieldset's name/value when child checkboxes had different name attributes (e.g. Drupal's colors[red], colors[blue] bracket notation). The group separates logic for name/value when all checkboxes share the same name VS each checkbox submits its own name/value independently.

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1626 🎟️


esteinborn:
@NovaCat35 for this fix, I'd like to attempt an alpha release for @loopy3025 to see if this works before merging to develop. Lets meet up to discuss how to release this as an alpha that Adam can test.

esteinborn:
@loopy3025 Can you test out the latest Alpha version where we tried to fix this for you:

`npm i @nysds/components@1.19.3-alpha-1`

loopy3025:
Hi there @NovaCat35 @esteinborn !  I didn't notice the alpha version. Woops! I jumped in and did a manual install from the working branch and it seemed to work great!

Manual build of the checkbox component from the `nysds-nys-checkbox` branch
Screenshot 2026-07-01 at 4 22 25 PM

I removed the checkboxgroup patch from my code.

Screenshot 2026-07-01 at 4 22 09 PM

The javascript no longer seems to be applying a name the the fieldset

Screenshot 2026-07-01 at 4 21 33 PM

The value is no longer forced on the fieldset when I submit the form. Instead, Drupal is allowed to use the individual values of the checkboxes on the backend to make its own array as designed.

Screenshot 2026-07-01 at 4 19 37 PM

The validation still works great, so that wasn't broken by the change.

Screenshot 2026-07-01 at 4 21 08 PM

I'd say this is a success. When I upgrade the module to the version this is released in, I should be able to remove the checkboxgroup patch.

esteinborn:
@loopy3025 Thanks for confirming, this is prepped for next week's release.

Issue #1412, opened 2026-03-26
[A11Y] UNav Header Translate lacks locale, needs Dropdown


mendeza:
The UNav Header has a "Translate" subcomponent with a list of available locales.

1. The component interaction lacks keyboard and screen reader affordances: refactor to use DS Dropdown
2. The locale items lack the `lang` attribute and appropriate values, providing a sub-par screen reader experience: update each item accordingly.

Issue #1093, opened 2025-12-05
[A11y Review] Avatar


mendeza:
## [Avatar](https://designsystem.ny.gov/components/avatar/)

plasticmind:
Status after #1632 (`fix/a11y-small-fixes`, commit `4e2d8383`):

**Fixed** — the captive default name: no `ariaLabel` now means decorative (hidden from AT) instead of announcing "avatar"; whitespace-only and `&nbsp;` values count as no name; an interactive avatar without a name logs a console warning.

**Remaining (this issue stays open for):** interactive mode is still a fake button (`role="button"` + `tabindex="0"`, no Enter/Space keydown handling). The #1466 recommendation stands — rebuild interactive mode on `nys-button circle` under the hood rather than extending the fake-button path.

Issue #1090, opened 2025-12-05
[A11y Review] Accordion


mendeza:
# [Accordion](https://designsystem.ny.gov/components/accordion/)

## Jesse's audit with Claude on 2026-01-29

>Content role="region" not linked to heading via aria-labelledby

---

## Vision
- [x] NVDA (Ethan, Aaron)
- [ ] JAWS (unclaimed)
- [ ] VoiceOver, macOS (Aaron)
- [ ] VoiceOver, iOS (Ethan)

## Mobility
- [ ] Dragon Naturally Speaking (Aaron)
- [ ] macOS Voice Control (Aaron)
- [ ] Windows 11 Voice Access (unclaimed)
- [ ] Keyboard (unclaimed)

---
FYI @EthanLigon @mendeza @schugartl

mendeza:
### Personae/AT
- Vision
- VoiceOver, iOS mobile

### Remarks
placeholder

### Video

Issue #1089, opened 2025-12-05
[SPIKE] Component inventory accessibility testing


EthanLigon:
### Spike Ticket Guidance

This is a test stub where a real issue would be. This is only a test.

PR #1632, opened 2026-06-23
feat(a11y): ElementInternals accessibility layer + WCAG 2.2 AA remediation across all components

### LINKED ISSUE #1092: [A11y Review] Alert
## [Alert](https://designsystem.ny.gov/components/alert/)

### LINKED ISSUE #1094: [A11y Review] Back To Top
## [Back To Top](https://designsystem.ny.gov/components/backtotop/)

_Let's take it from the top! One, two, three, four..._

### LINKED ISSUE #1099: [A11y Review] File Input
# [File Input](https://designsystem.ny.gov/components/fileinput/)

This issue is a stub.

## Jesse's audit with Claude on 2026-01-29

>Dropzone not keyboard accessible — no keydown handler

### LINKED ISSUE #1101: [A11y Review] Global Header
[Global Header](https://designsystem.ny.gov/components/globalheader/)

### LINKED ISSUE #1103: [A11y Review] Modal
# [Modal](https://designsystem.ny.gov/components/modal/)

This issue is a stub.

## Jesse's audit with Claude on 2026-01-29

>Dismiss button aria-label set to " " for 100ms before update

### LINKED ISSUE #1104: [A11y Review] Pagination
# [Pagination](https://designsystem.ny.gov/components/pagination/)

### LINKED ISSUE #1108: [A11y Review] Stepper
# [Stepper](https://designsystem.ny.gov/components/stepper/)

This issue is a stub.

## Jesse's audit with Claude on 2026-01-29

>Contradictory aria-hidden="true" AND role="button" on step labels

## Findings

* Step buttons lack name, role, state
* List lacks semantics (is a list, number of items)
* Must clean up invalid "button" code
  * if interactive, use actual button
  * do not use aria-hidden
  * do not use tabindex at all
  * Button name must contain the step number; component could be refactored to simplify structure and include number and visually-hidden "step" text all in one button

## Static display scenario
* when step headers are display only 
  * either header is repeated above content, ideally with extra info about number of step
  * or if no header repeated, content is aria-labelledby the step

## Action/navigation scenario
* when step headers are links/buttons then set role navigation
* keep current step interactive to allow valid use of aria-current="step"
* when activation causes dynamic change to partial page content, ensure focus stays on last step activated (realistic example suggested, also call out in guidance)

### LINKED ISSUE #1114: [A11y Review] UNav Header
## [UNav Header](https://designsystem.ny.gov/components/unav-header/)

### LINKED ISSUE #1355: A11Y: Stepper fails automated test
@hemanthkuvvarapu [reports ](https://teams.microsoft.com/l/message/19:l5Zo00jgzpzTb5ACeXQQVatofIuRI-C0ThuQXbndb9U1@thread.tacv2/1772136580561?tenantId=f46cb8ea-7900-4d10-8ceb-80e8c1c81ee7&groupId=40dc9e8f-13b9-4301-8cb7-db5c2b21c9fa&parentMessageId=1772136580561&teamName=NYS%20Design%20System&channelName=General&createdTime=1772136580561) that the Stepper in his project fails Chrome Lighthouse automated testing with:
```
[aria-hidden="true"] elements contain focusable descendents.
```

Accessibility Insights likewise fails with:

<strong>aria-hidden-focus</strong></span>: <span class="rule-details-description">Ensure aria-hidden elements are not focusable nor contain focusable elements</span></span>

<table class="report-instance-table--3VoZX"><tbody><tr class="row---3i35"><th class="row-label--FlVZl">Path</th><td class="row-content--07KUh">#nys-stepper-1-1772820033856 &gt; nys-step[label="Personal Details"][first=""][previous=""],.nys-step__content[aria-hidden="true"]</td></tr><tr class="row---3i35"><th class="row-label--FlVZl">Snippet</th><td class="row-content--07KUh snippet--hSNfv">&lt;div class="nys-step__content" tabindex="-1" aria-hidden="true"&gt;
            &lt;div class="nys-step__label" role="button" aria-hidden="true" aria-label="Personal Details Step" tabindex="0"&gt;
              &lt;!--?lit$628770297$--&gt;Personal Details
            &lt;/div&gt;
          &lt;/div&gt;</td></tr><tr class="row---3i35"><th class="row-label--FlVZl">Related paths</th><td class="row-content--07KUh"><ul class="path-list--MxxKM"><li>#nys-stepper-1-1772820033856 &gt; nys-step[label="Personal Details"][first=""][previous=""],.nys-step__label[aria-label="Personal Details Step"][role="button"]</li></ul></td></tr><tr class="row---3i35"><th class="row-label--FlVZl">How to fix</th><td class="row-content--07KUh"><div class="how-to-fix-content--Zz96u"><div><div>Fix the following:</div><ul><li>Focusable content should have tabindex="-1" or be removed from the DOM</li></ul></div></div></td></tr></tbody></table>

#### References

* https://accessibilityinsights.io/info-examples/web/aria-hidden-focus/
* https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html

### LINKED ISSUE #1694: FIX: nys-toggle `prefers-reduced-motion` override is dead code — animation never disabled [NEW — v1.19.2]
**Labels:** `bug`, `accessibility`, `high`
**Effort:** Minimal (15 minutes)

**Description:**

`nys-toggle.scss:20` defines and `:180` consumes `--_nys-toggle-transition-duration: 0.3s`. The `@media (prefers-reduced-motion: reduce)` block at `:230-234` sets a *differently named* property — `--toggle-transition-duration: 0s` (missing the `_nys-` segment) — so the override has no effect. Toggles animate at full duration for users who requested reduced motion. Present identically at the v1.18.1 tag; previously mis-filed as a naming-convention item (v1 Issue 20). Outside PR #1632's scope.

**Affected file:** `packages/nys-toggle/src/nys-toggle.scss:232`

**Acceptance criteria:**
- [ ] Line 232 changed to `--_nys-toggle-transition-duration: 0s;`
- [ ] Verified under emulated `prefers-reduced-motion: reduce` that the transition computes to 0s
- [ ] Repo swept for other set-but-never-consumed custom-property overrides

### LINKED ISSUE #1726: FIX: Global header - clones nav links into shadow DOM and forces pathname-based active state, ignoring author `aria-current`
## Summary
`nys-globalheader` **clones** the author's slotted nav links into its shadow DOM (for both the desktop and mobile copies) and then applies its **own `window.location.pathname`-based** active-link logic, ignoring any `aria-current` the author set. In apps that don't use pathname routing (e.g. hash-routed SPAs), the wrong link is marked active, and the author's `aria-current` is overridden.

## Component / Area
`nys-globalheader`

## Current behavior
- Slotted nav nodes are cloned into a shadow-DOM container: `container!.appendChild(node.cloneNode(true))` — `packages/nys-globalheader/src/nys-globalheader.ts:138`.
- Active state is computed from the URL path: `const currentUrl = window.location.pathname.replace(/\/+$/, "") || "/"` — `:82`; the component clears all `active` classes (`:106-107`) and sets `active` on its own best pathname match (`:110`).
- There is no code path that reads or honors an author-provided `aria-current` on the links; the component is the sole authority on which link is "active."

## Expected behavior
Respect an author-supplied `aria-current` (or another explicit opt-out) instead of unconditionally overriding it with pathname matching. Pathname matching is a reasonable default, but it should defer to an explicit author signal so hash-routed / non-pathname apps can mark the correct active link. Cloning the links into the shadow DOM also means author state/handlers on the originals don't reach the rendered copy.

## Steps to reproduce
1. Use `nys-globalheader` with slotted `<ul><li><a>` nav links in a hash-routed (or otherwise non-pathname) app.
2. Set `aria-current="page"` on the link that is actually active.
3. The component marks a *different* link active based on `window.location.pathname`, and the author's `aria-current` is not reflected.

## Notes
- The overridden `aria-current` is an assistive-tech concern (screen readers announce the wrong "current page").
- Related: **#889** ("`nys-globalheader` & `data-active` for active link") tracks giving authors control over the active link via a `data-active` attribute; this bug is the same underlying need surfaced as an `aria-current` override + shadow-DOM clone defect.
- Same "clone slotted content into shadow DOM" family as `nys-table` (#1650) and `nys-tabgroup`.

### LINKED ISSUE #1763: a11y: nys-fileinput error is never associated with the exposed control
## Problem

`nys-fileinput`'s native `<input type="file">` carries `hidden`, `aria-hidden="true"` and `tabindex="-1"`, so it is **not** the control in the accessibility tree. But that hidden input is where `aria-invalid` and `aria-errormessage` are set:

```
packages/nys-fileinput/src/nys-fileinput.ts:556-558
  aria-invalid=${this.showError ? "true" : "false"}
  aria-errormessage=${this.id + "--error"}
  aria-hidden="true"
```

The control a user actually reaches is the `nys-button` (or the dropzone). Its description is a hand-built string (`_buttonAriaDescription`), assembled separately from `<nys-errormessage>`. So the error association points at an element no assistive technology can see, and the real control's error state depends on string plumbing rather than an ARIA relation.

## Impact

A screen-reader user focused on the "Choose file" button gets whatever `_buttonAriaDescription` happens to contain. There is no `aria-invalid` on the exposed control and no relation to the rendered `<nys-errormessage>`.

## Suggested fix

Move `aria-invalid` and the error association onto the element that is actually exposed (the `nys-button` / dropzone), referencing the `<nys-errormessage>` element rather than duplicating its text into a string.

Note the related constraint found while fixing the other form controls: Chromium does not surface `aria-errormessage` for a control inside a shadow root, so the association must go through `aria-describedby`. See `scripts/verify-a11y-names.mjs`.

## Context

Found while implementing cross-shadow label/error association. Deliberately left out of that branch — it needs a design decision about which element is the control, not a mechanical fix.

### LINKED ISSUE #1764: a11y: nys-fileitem passes error text where aria-errormessage expects an id
## Problem

`nys-fileitem` passes the error **message text** into `aria-errormessage`, which takes an **IDREF**:

```
packages/nys-fileinput/src/nys-fileitem.ts:96-97
  aria-invalid="true"
  aria-errormessage=${this.errorMessage}
```

`aria-errormessage` must name the `id` of an element containing the error. Given text like `"File too large"`, the browser looks for an element with `id="File too large"`, finds nothing, and the relation silently resolves to nothing.

## Impact

The error is never conveyed through the ARIA relation. `aria-invalid="true"` is also hardcoded, so the item is announced as invalid regardless of whether it actually is.

## Suggested fix

Render the error in an element with a stable id and reference that id, matching the pattern used by the other form controls. Make `aria-invalid` reflect real state rather than a constant.

Note: Chromium does not surface `aria-errormessage` for a control inside a shadow root at all, so the association also needs `aria-describedby`. See `scripts/verify-a11y-names.mjs` and the fix applied to the other form controls.

## Context

Found while implementing cross-shadow label/error association; left out of that branch as a separate defect.

### LINKED ISSUE #1794: FIX: a11y: `nys-button` does not forward `aria-expanded`/`aria-controls` to the shadow-root button
## Summary

ARIA state attributes set on a `<nys-button>` host element never reach the `<button>` inside its shadow root, so disclosure state is not exposed to assistive technology. This is currently live on the design system site via `nys-unavheader`.

A custom element with no implicit or explicit `role` does not map its ARIA attributes into the accessibility tree. `aria-expanded` on the host is therefore ignored, and the real `<button>` in the shadow root never receives it.

## Reproduction

designsystem.ny.gov (NYSDS 1.19.3), any browser. In the console:

```js
const host  = document.querySelector('nys-unavheader')
                .shadowRoot.querySelector('#nys-unavheader__know--inline');
const inner = host.shadowRoot.querySelector('button');

host.getAttribute('aria-expanded');   // "false"
inner.getAttribute('aria-expanded');  // null   ← never set

inner.click();                        // expand the trust bar

host.getAttribute('aria-expanded');   // "true"  ← host updates correctly
inner.getAttribute('aria-expanded');  // null   ← still never set
```

The host reflects state correctly. The element that actually carries the `button` role does not.

## Impact

A screen reader announces *"Here's how you know, button"* with no collapsed/expanded state, and activating it announces no state change. The control gives no indication it is a disclosure.

Fails **WCAG 2.2 SC 4.1.2 Name, Role, Value (Level A)** — states must be programmatically determinable.

Because the defect is in `nys-button` rather than in any one consumer, it affects every NYSDS consumer that uses `nys-button` as a disclosure trigger.

## Affected instances found on the homepage

All inside `nys-unavheader`; in each case the inner `<button>` has `aria-expanded: null`:

| Host id | Host `aria-expanded` | Host `aria-controls` |
| --- | --- | --- |
| `nys-unavheader__closetrustbar` | `"false"` | — |
| `nys-unavheader__know--inline` | `"false"` | `nys-unavheader__closetrustbar` |
| `nys-unavheader__searchbutton` | `"false"` | — |

`aria-controls` is dropped by the same mechanism.

## Suggested fix

In `nys-button`, forward author-supplied ARIA attributes to the internal `<button>` rather than leaving them on the host. Either:

- accept them as reactive properties and render them onto the internal button, observing changes so state updates propagate; or
- expose state from the host via `ElementInternals` / `ARIAMixin` (e.g. `this.#internals.ariaExpanded`), which maps host-level state into the a11y tree without the shadow-root hop.

A general solution covering arbitrary `aria-*` forwarding would also close the `aria-label` / `aria-describedby` request in #1464.

## Related

- #1464 — same underlying mechanism (author ARIA not reaching the shadow root), scoped to `aria-label` / `aria-describedby`. Worth fixing together; happy to see this folded in if the team prefers one issue.
- #1114 — `[A11y Review] UNav Header`, which notes the language picker lacking `aria-expanded`. This issue is the component-level root cause for the header's other disclosure buttons.
- #1089 — `[SPIKE] A11y status of components, build-test pages, and ref site`.

## Verification

Confirmed via axe-core 4.10.2 (`aria-allowed-attr`, flagged incomplete: *"Check that there is no problem if the ARIA attribute is ignored on this element: aria-expanded=false"*) plus the manual shadow-root inspection above, at 1280px and 320px viewports.

### LINKED ISSUE #1814: chore(a11y): migrate nys-iconlist and nys-processlist to @nysds/internals
Follow-up to #1632. Both components predate the internals layer and still hand-roll ID counters. Migrate both to NysElement (keep role attributes — axe cannot see ElementInternals roles on these light-DOM lists), wire @nysds/internals devDependency + tsconfig reference, delete processlist's three hand-rolled ARIA props (host is light DOM; native attributes already work), and add a dev-time warning when non-item children are silently hidden (WCAG 1.3.1 hazard).

### LINKED ISSUE #1815: fix(a11y): nys-verticalnav — migrate to @nysds/internals and fix critical accessible-name/mobile bugs
Follow-up to #1632. Migration: NysElement for both elements (also fixes a real ID-collision bug — nys-verticalnavgroup generates ids prefixed nys-verticalnav- from a separate counter), add missing sideEffects key. Critical fixes: (1) the nav landmark's aria-labelledby dangles when a consumer fills the header slot — the generated heading is slot fallback content, so the landmark has NO accessible name (WCAG 4.1.2/2.4.1); (2) the mobile template drops the header slot entirely (1.3.1); (3) mobile nav has no accessible name; (4) documented disabled-link pattern (a[aria-disabled] without href) is color-only — normalize with role=link + tabindex; (5) slotchange listener lost when crossing the 1024px breakpoint. Plus: keyboard test suite for nys-verticalnavgroup (currently none).

### LINKED ISSUE #1816: feat(a11y): nys-radiobutton cross-shadow labelling (labelledby/hideLabel)
Follow-up to #1632 / #1765. Radiobutton was deliberately left out of cross-shadow labelling pending #1721, which closed 2026-08-03. Add the labelledby + hideLabel props mirroring nys-checkbox's implementation. Note: radiobutton is now a light-DOM component (no shadow root), so external label association can use plain document-scope IDREFs — simpler than the checkbox's associateControlRefs path; verify with the CDP computed-name harness.

### LINKED ISSUE #1817: FIX: a11y: inert ariaLabel/ariaDescription attributes leave icon-only buttons unnamed (repo-wide sweep)
## Overview

Successor to #1464 (closed) for the remaining repo-wide work, surfaced during #1632.

`nys-button` has no `ariaLabel`/`ariaDescription` props, so components that set those attributes on it are setting attributes that reach nothing — leaving **unnamed icon-only buttons**:

- `nys-fileitem` — remove button
- `nys-combobox` — clear and toggle buttons
- `nys-alert` — close button
- `nys-checkbox`
- `nys-pagination` — all buttons

#1632 fixed `nys-unavheader`'s instances using the `label`/sr-only slot pattern, but the pattern was never swept repo-wide.

## Fix

Either sweep each component onto the `label`/sr-only slot pattern used in unavheader, or add a real `ariaLabel` prop to `nys-button` and route it properly. Verify each button's computed accessible name.

## Acceptance criteria

- No icon-only `nys-button` instance in the packages above has an empty computed accessible name
- Tests assert the accessible name (not just attribute presence)

Merges into #1632 (`fix/accessibility-remediation`) via branch `fix/a11y-arialabel-sweep`.

### LINKED ISSUE #1818: BUG: a11y: nys-accordion needs author-settable heading levels and region naming
## Overview

Follow-up to the audit finding in #1090 and work landed in #1632. The accordion trigger is not wrapped in a real heading with an author-settable level, and its `role="region"` naming semantics are incomplete.

This is not just an accordion problem: it **caps `nys-verticalnav`'s mobile heading semantics** (see #1815), which renders its mobile disclosure through the accordion pattern and inherits the missing heading level.

## Fix

- Add a `headingLevel` (or equivalent) prop so authors can place the accordion trigger in the correct document outline position (`<h2>`–`<h6>` wrapping the trigger button)
- Ensure each panel's `role="region"` is labelled by its trigger and only applied where appropriate (consider dropping `region` when many panels would flood the landmark list)
- Let `nys-verticalnav` mobile pass its heading level through

## Acceptance criteria

- Author can set heading level; document outline is correct in AT
- Panels have correct accessible names; axe clean
- verticalnav mobile heading semantics no longer limited by accordion

Merges into #1632 (`fix/accessibility-remediation`) via branch `fix/a11y-accordion-headings`.

### LINKED ISSUE #1819: FIX: a11y: sweep components that render nys-* children without importing them (vacuous axe audits)
## Overview

Surfaced during #1632. Components that render `nys-*` children in their templates **without importing those child packages** leave the children as inert unknown elements when the component is used in isolation. Worse, their axe audit tests **pass vacuously** — an unupgraded custom element has no semantics for axe to flag, so the audit silently tests nothing.

This bit `nys-unavheader` (fixed in #1632); the rest of the repo is unaudited for the pattern.

## Fix

- Sweep all packages: any component whose template renders another `nys-*` element must import it (with honest `sideEffects` declarations, following the pattern established in #1632)
- Add a guard against regression — e.g. a test-time check that all rendered `nys-*` children are defined (`customElements.get()`), or a lint/CI rule

## Acceptance criteria

- Every component upgrades all of its rendered `nys-*` children when used standalone
- Audit tests fail (not pass) when a child element is not upgraded

Merges into #1632 (`fix/accessibility-remediation`) via branch `fix/a11y-self-registration-sweep`.

### LINKED ISSUE #1820: FIX: a11y: standalone nys-radiobutton builds a synthetic aria-label instead of IDREF association
## Overview

Small remainder from #1632, distinct from #1816 (which adds `labelledby`/`hideLabel` for cross-shadow labelling and lands with that PR). When `nys-radiobutton` is used standalone with its own visible label, it still builds a **synthetic `aria-label`** for the input rather than associating the visible label via IDREF (`aria-labelledby`/`for`).

Synthetic labels violate the label-in-name expectation (WCAG 2.5.3) risk and drift from the visible text. Radiobutton is now light-DOM, so plain document-scope IDREFs work — same reasoning as the `nys-textinput` fix landed in #1632.

## Fix

Associate the visible label via IDREF; delete the synthetic `aria-label` path. Verify the computed accessible name matches the visible label text.

Merges into #1632 (`fix/accessibility-remediation`) via branch `fix/a11y-small-fixes` (grouped with other small a11y remainders: #1101 mobile focus trap, #1093 avatar default name, #1795 landmark label overridability).

### LINKED ISSUE #1821: FIX: nys-card predates the @nysds/internals patterns (#1632) — migrate to NysElement
## Overview

nys-card (#1813) landed on develop in parallel with #1632, so it's the one component not on the shared accessibility foundation that PR establishes:

- `extends LitElement` directly instead of `NysElement`
- Hand-rolled id counter (`nys-card-${Date.now()}-${counter++}`) instead of the base class's `generateId` format
- `"sideEffects": true` instead of the honest sideEffects array pattern
- No `@nysds/internals` devDependency or tsconfig reference (child-component references were already added in the #1632 develop sync)

Also worth review while in there: the clickable-card modes (`href`, `onClick`, `target`) — verify the interactive surface is a real link/button with a correct accessible name, per the no-fake-buttons rule.

Merges into #1632 (`fix/accessibility-remediation`) via branch `fix/a11y-card-internals` — held until the current review round completes.

plasticmind:
## Summary

Integrates **ElementInternals as the accessibility-semantics layer** across all ~39 `nys-*` components, backed by a new shared `@nysds/internals` workspace module, and remediates a set of concrete WCAG 2.2 AA gaps surfaced by a full component audit. No breaking changes to public component API.

The audit reframed the original premise: form-association (`formAssociated` + `attachInternals` for `setFormValue`/`setValidity`) already existed on the form controls. The real gap was that ElementInternals was **not** used as the accessibility layer — default roles and ARIA state were hand-wired per component (or missing), there was no shared infrastructure, and a few real WCAG gaps existed (e.g. `nys-textinput` building a synthetic duplicated `aria-label` instead of associating its visible `<nys-label>`).

### New shared layer — `packages/internals` (`@nysds/internals`)
A private, unpublished workspace module (consumed as a `devDependency`, inlined at build) providing a layered, tree-shakeable Lit mixin architecture:

- `NysElement` / `IdentifiedMixin` — auto id generation only (presentational components that don't reflect a host role)
- `NysReflectsAriaElement` / `ReflectsAriaMixin` — adds host `role` + ARIA-state reflection via `ElementInternals` (currently `nys-divider` → `role="separator"`)
- `NysFormControlElement` / `FormControlMixin` — adds form association, validity helpers (`setFormValue`, `setValidityFromState`, `clearValidity`), and host `aria-invalid`

Plus à-la-carte ARIA-association helpers (`associateControl`/`associateHost` with element-reference support + IDREF fallback) and `generateId` (preserves the legacy id format). 25 unit tests across 6 browsers.

### Component migrations
- **10 form controls** → `NysFormControlElement`: removed per-component `_internals`/`formAssociated`/id-counter boilerplate; route value/validity through the shared helpers.
- **~26 presentational components** → `NysElement` (id-only); `nys-divider` → `NysReflectsAriaElement`.

### The hybrid accessibility approach (intentional, per component)
- **ElementInternals AOM** for form value/validity, host `aria-invalid`, and host roles where the host *is* the semantic element (`nys-divider`).
- **IDREF / template ARIA** for accessible *names* on inner native controls (`aria-labelledby` → the visible `<nys-label>`, replacing synthetic `aria-label`) and for inner/landmark roles (alert live region, breadcrumbs/pagination `nav`, tabs). This is deliberate: ElementInternals reflects on the *host's* a11y node, not the inner `<input>` the AT interacts with under `delegatesFocus`.

### WCAG fixes landed
`nys-textinput` visible-label association (1.3.1 / 2.5.3); `nys-datepicker` error text associated via `aria-errormessage` + `aria-invalid`; `nys-breadcrumbs` / `nys-pagination` `aria-current` + `nav` landmark; `nys-dropdownmenu` menu accessible name; `nys-accordion` heading/region; `nys-avatar` name + decorative icon; `nys-backtotop` reduced-motion handling; and others.

### Reliability + tooling
- Form controls now **self-register** their internal `<nys-label>`/`<nys-errormessage>` (with honest `sideEffects`) so the accessible-name association is reliable.
- plop templates scaffold new components onto the correct base + dep + `sideEffects` + tsconfig reference.
- MCP form-validation guide corrected (`aria-errormessage`/`aria-labelledby`, not `aria-describedby`).
- Build-order, `tsconfig.build.json`, and CEM config wired for the new package (excluded from the manifest — it declares no elements).

## Notes for reviewers
- **No public reactive props live in the mixins** — they'd vanish from the Custom Elements Manifest (and thus React wrappers/MCP/Storybook controls). All `@property` declarations stay on the leaf components; the CEM diff confirms public attributes/props/events are unchanged.
- Concrete base classes are exported as `export const X = /*@__PURE__*/ Mixin(LitElement)` (not `class X extends Mixin(...)`) so unused mixins tree-shake out of each bundle.

## Verification
- 37/37 packages build cleanly; ESLint clean.
- **1011 tests pass across all 6 browser targets**; coverage 91.4% statements / 86.9% functions / 80.5% branches (above the 80% gate).
- Branch is synced to the latest `develop` (merge resolved: kept the `aria-errormessage` remediation over the older `aria-describedby` pattern, combined `nys-backtotop` SSR-guard + reduced-motion, aligned versions to 1.19.1, deduped `sideEffects` keys, regenerated the manifest + React wrappers).

The full PRD lives in-repo at `docs/accessibility-remediation-PRD.md`.

## Out of scope (tracked separately)
- `nys-label`/`nys-errormessage` standalone-vs-internal architecture refactor.
- `packages/` reorganization into `components`/`foundations`/`frameworks`/`tools`.

---

## Linked issues (updated 2026-08-05)

This branch now also carries the develop resync plus nine follow-up ticket branches (one merge commit each — reviewable per ticket).

### Closes

Closes #1094
Closes #1355
Closes #1694
Closes #1726
Closes #1763
Closes #1764
Closes #1794
Closes #1814
Closes #1815
Closes #1816
Closes #1092
Closes #1099
Closes #1101
Closes #1103
Closes #1104
Closes #1108
Closes #1114
Closes #1817
Closes #1818
Closes #1819
Closes #1820
Closes #1821

> ⚠️ **Merge note:** GitHub only auto-closes on merges to the default branch (`main`). This PR targets `develop`, so these will *link* but not close at merge — either close the list manually when this merges, or carry the keywords into the next develop→main release PR.

### Addresses (partial — verified against code, intentionally NOT auto-closing)

- #1089 — umbrella tracker; many component rows advanced by this PR.
- #1093 (Avatar) — captive default name removed (unnamed = decorative, whitespace/`&nbsp;` rejected, dev warning for unnamed interactive avatars); the fake-button-to-`nys-button` refactor for interactive mode remains.
- #1090 (Accordion) — audit finding fixed; heading semantics now done via #1818 (real `<h2>`–`<h6>` triggers, author-settable `headingLevel`). AT test matrix remains.
- #1412 (Translate) — locale tagging, operability, and now full APG menu-button semantics with arrow-key navigation done; the DS Dropdown refactor it asks for was intentionally not done here.
- #1795 (Landmarks) — all four paired landmarks have distinguishing accessible names, now author-overridable via a consistent `landmarkLabel` prop; axe's count-based duplicate-landmark rules still fire on paired pages (needs a landmark-demotion decision).

### Follow-up fixes (2026-08-06)

The four issues surfaced during this work were filed (#1817 inert `ariaLabel`/`ariaDescription` sweep, #1818 accordion heading levels, #1819 self-registration sweep, #1820 standalone radiobutton labelling) and are fixed in this branch, alongside the remaining findings of #1092, #1099, #1101, #1103, #1104, #1108, and #1114 — ten ticket branches, one reviewable merge commit each:

- `fix/a11y-arialabel-sweep` (#1817) — every icon-only `nys-button` named via `label`/sr-only; dead `ariaLabel`/`ariaDescription` bindings removed repo-wide (fileitem, combobox, alert, pagination, table, tooltip, fileinput; checkbox needed nothing)
- `fix/a11y-pagination` (#1104) — `ariaCurrent` forwarded to the inner button (new `nys-button` prop); three focus-management bugs fixed; ellipsis no longer a fake tab stop
- `fix/a11y-accordion-headings` (#1818) — real `<h2>`–`<h6>` triggers with group + per-item `headingLevel`; verticalnav mobile passes its level through
- `fix/a11y-self-registration-sweep` (#1819) — ~24 packages now import the `nys-*` children they render; `findUnregisteredChildren` test guard prevents vacuous axe audits; `build-order.js` topologically sorted; tsconfig project references made honest in 26 packages
- `fix/a11y-modal-arialabel` (#1103) — `ariaLabel` setTimeout hack replaced with declarative `aria-labelledby`; `role="dialog"`/`aria-modal` moved onto the element that receives focus; close button actually named
- `fix/a11y-alert-live-region` (#1092) — `role`/`aria-live`/`aria-atomic` consolidated on the content wrapper; assertive/polite split by type
- `fix/a11y-fileinput-dropzone` (#1099) — dropzone hint text associated with the real button; dead `aria-label` removed from the non-interactive wrapper
- `fix/a11y-unav-language-menu` (#1114) — APG menu-button semantics with roving tabindex, arrow keys, Home/End, Escape, and click-away close
- `fix/a11y-stepper-buttons` (#1108) — real `<button>` steps in `<ol>`/listitem semantics; non-navigable steps genuinely inert
- `fix/a11y-small-fixes` — global header mobile focus trap (#1101), avatar captive default name (#1093), standalone radiobutton IDREF labelling (#1820), author-overridable landmark labels (#1795)

Verified: full test suite across 6 browsers, `build:all` from clean, lint/stylelint clean, and a manual VoiceOver pass over all changed components.

Late addition after the develop re-sync: `fix/a11y-card-internals` (#1821) migrates the newly-landed nys-card onto NysElement + the internals patterns, fills its Code Connect mapping, and aligns its version. The clickable-card heading-flattening finding is documented on #1821 for design triage.

**New follow-ups surfaced (not yet filed):** modal fires a duplicate `nys-close` per dismissal; modal `width` prop is inert; stepper's mobile counter toggle is still a `div role="button"`; fileinput's state-aware button descriptions were removed as dead code and need a real implementation; a page-level live-region pattern for fully-formed alert injection; unavfooter's inner nav label doesn't follow `landmarkLabel`.


esteinborn:
currently failing to build any packages:
image

plasticmind:
@esteinborn Good catch. I'm going to assume you got that when running locally. Turns out there's a `tsc` build cache that gets put into the component folder. It's .gitignored, but can bite you with a new shared package (like the new @internals one). That's why it wouldn't show up when building in a new environment, but I've updated `clean:dist` to clean these out so that we shouldn't have this problem going forward on local.

To see this in action, pull the latest and run ` npm run clean:dist && npm run build:packages`

packages/nys-breadcrumbs/src/nys-breadcrumbs.ts
```diff
@@ -223,6 +222,9 @@ export class NysBreadcrumbs extends LitElement {
     liEl.className = "nys-breadcrumbitem";
 
     if (isCurrentPage) {
+      // WCAG 4.1.2 / WAI-ARIA breadcrumb pattern: the current page must be
+      // programmatically identified so assistive tech announces "current page".
+      liEl.setAttribute("aria-current", "page");
       liEl.textContent = label;
       liEl.setAttribute("aria-current", "page");
```
esteinborn: `liEl.setAttribute("aria-current", "page");`

Twice?

packages/nys-datepicker/src/nys-datepicker.ts
```diff
@@ -1,6 +1,13 @@
-import { LitElement, html, nothing, unsafeCSS } from "lit";
+import { LitElement, html, unsafeCSS } from "lit";
```
esteinborn, line 1: you replaced LitElement in other components, shoudl this be removed as well?

packages/nys-globalfooter/src/nys-globalfooter.ts
```diff
@@ -45,6 +50,12 @@ export class NysGlobalFooter extends LitElement {
    * --------------------------------------------------------------------------
    */
 
+  // super.connectedCallback() (NysElement) assigns an id when one is
+  // not provided. The contentinfo landmark intentionally stays on the inner
+  // <footer> element (and is given an accessible name from the agency name so
+  // multiple footers on a page are distinguishable), so this component keeps
+  // defaultRole = null and does not move a role onto the host.
+
```
esteinborn: this comment references `super.connectedCallback()` but then doesnt run that function. should it?

packages/nys-select/src/nys-select.ts
```diff
@@ -1,12 +1,17 @@
 import { LitElement, html, unsafeCSS } from "lit";
```
esteinborn, line 1: Do we still need LitElement here?

packages/nys-textarea/src/nys-textarea.ts
```diff
@@ -1,11 +1,16 @@
 import { LitElement, html, unsafeCSS } from "lit";
```
esteinborn, line 1: Do we still need LitElement?

packages/nys-textinput/src/nys-textinput.ts
```diff
@@ -1,11 +1,16 @@
 import { LitElement, html, unsafeCSS } from "lit";
```
esteinborn, line 1: still need LitElement?

packages/nys-video/src/nys-video.scss
```diff
@@ -148,22 +161,26 @@
 
         &:focus-visible {
           outline: 3.5px solid var(--_nys-video-outline-color--play-button);
-          outline-offset: 1px;
+          outline-offset: -3.5px;
```
esteinborn: Shouldn't this be a variable?

packages/nys-video/src/nys-video.ts
```diff
@@ -87,26 +86,19 @@ export class NysVideo extends LitElement {
...
 
     if (this.autoplay) {
       this._announceVideoVO();
     }
   }
 
-  disconnectedCallback() {
-    super.disconnectedCallback();
-  }
-
```
esteinborn, line 171: Why did we remove the disconnectedCallback?

esteinborn (changes requested):
A few questions but otherwise this looks good. Pulling and testing now.

esteinborn:
Also noting here that this broke accordion display:
image

packages/internals/src/form-control-mixin.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    checkValidity(): boolean {
+      return this.internals?.checkValidity() ?? true;
+    }
+
+    reportValidity(): boolean {
+      return this.internals?.reportValidity() ?? true;
+    }
+
+    formResetCallback(): void {
```
NovaCat35, line 78: The `formResetCallback()` only clears validity, not value/state. Furthermore, in some components, this feature differs case by case due to the different props a component have

packages/internals/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@nysds/internals",
+  "version": "1.19.1",
```
NovaCat35: Version needs update

packages/nys-accordion/src/nys-accordionitem.ts
```diff
@@ -136,29 +132,58 @@ export class NysAccordionItem extends LitElement {
 
   render() {
     const contentId = `${this.id}-content`;
+    const buttonId = `${this.id}-button`;
+    // Clamp the heading level to the valid ARIA range (1-6).
+    const level = Math.min(6, Math.max(1, this.headingLevel || 3));
```
NovaCat35: I question the need for this level variable. Do we really need VO to announce "Heading Level #..." in accordion items?

packages/nys-accordion/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-icon": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-alert/package.json
```diff
@@ -27,6 +27,7 @@
     "@nysds/nys-button": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-avatar/package.json
```diff
@@ -25,6 +25,7 @@
     "@nysds/nys-icon": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-backtotop/package.json
```diff
@@ -27,6 +27,7 @@
     "@nysds/nys-button": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-badge/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-icon": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-breadcrumbs/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-icon": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-button/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-icon": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-checkbox/package.json
```diff
@@ -28,6 +28,7 @@
     "@nysds/nys-errormessage": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-combobox/package.json
```diff
@@ -22,7 +22,12 @@
     "test:watch": "vite build && wtr --watch",
     "lit-analyze": "lit-analyzer 'src/**/*.ts'"
   },
+  "dependencies": {
+    "@nysds/nys-label": "^1.19.2",
+    "@nysds/nys-errormessage": "^1.19.2"
+  },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-datepicker/package.json
```diff
@@ -30,6 +30,7 @@
     "wc-datepicker": "^0.10.0"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-divider/package.json
```diff
@@ -23,6 +23,7 @@
     "lit-analyze": "lit-analyzer '**/*.ts'"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-dropdownmenu/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-icon": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-fileinput/package.json
```diff
@@ -29,6 +29,7 @@
     "@nysds/nys-errormessage": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-globalfooter/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-divider": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-globalheader/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-icon": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-icon/package.json
```diff
@@ -23,6 +23,7 @@
     "dist/"
   ],
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-modal/package.json
```diff
@@ -25,6 +25,7 @@
     "@nysds/nys-button": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-pagination/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-button": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-radiobutton/package.json
```diff
@@ -27,6 +27,7 @@
     "@nysds/nys-errormessage": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-select/package.json
```diff
@@ -27,6 +27,7 @@
     "@nysds/nys-errormessage": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-skipnav/package.json
```diff
@@ -23,6 +23,7 @@
     "lit-analyze": "lit-analyzer '**/*.ts'"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-stepper/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-button": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-tab/package.json
```diff
@@ -26,6 +26,7 @@
     "@nysds/nys-button": "1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-table/package.json
```diff
@@ -27,6 +27,7 @@
     "@nysds/nys-button": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-textarea/package.json
```diff
@@ -28,6 +28,7 @@
     "@nysds/nys-errormessage": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-textinput/package.json
```diff
@@ -27,6 +27,7 @@
     "@nysds/nys-errormessage": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-toggle/package.json
```diff
@@ -22,9 +22,11 @@
     "lit-analyze": "lit-analyzer '**/*.ts'"
   },
   "dependencies": {
-    "@nysds/nys-icon": "^1.19.2"
+    "@nysds/nys-icon": "^1.19.2",
+    "@nysds/nys-label": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-tooltip/package.json
```diff
@@ -23,6 +23,7 @@
     "lit-analyze": "lit-analyzer '**/*.ts'"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-unavfooter/package.json
```diff
@@ -23,6 +23,7 @@
     "lit-analyze": "lit-analyzer '**/*.ts'"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

packages/nys-unavheader/package.json
```diff
@@ -28,6 +28,7 @@
     "@nysds/nys-textinput": "^1.19.2"
   },
   "devDependencies": {
+    "@nysds/internals": "^1.19.1",
```
NovaCat35: Version needs update

NovaCat35 (changes requested):
I must be really bad at Q&A or this PR is very good. After much testing in React Demo and scouring some logic of the new PR codebase, I'm happy to not find heavy breaking changes

packages/nys-accordion/src/nys-accordionitem.ts
plasticmind: That's the [WAI recommendation](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#wai-ariaroles,states,andproperties)

packages/nys-breadcrumbs/src/nys-breadcrumbs.ts
plasticmind: Good catch. Fixed.

mendeza:
Regarding proposed Accordion changes,  please see https://github.com/ITS-HCD/nysds/issues/1818#issuecomment-5208422866

mendeza:
Regarding Checkbox changes, please see https://github.com/ITS-HCD/nysds/issues/1818#issuecomment-5218666502

PR #1635, opened 2026-06-24
Fix: Accordion height to recalculate for bordered and new content expansion

### LINKED ISSUE #1633: BUG: Bordered `nys-accordionitem` expanded by default clips bottom border (height measured before border applied)
## Summary

When a **bordered** `nys-accordionitem` is rendered `expanded` by default, the bottom of the panel (including the bottom border) is clipped. The JS height calculation does not account for the border that the parent applies to the content container, so the fixed pixel height locked onto the `overflow: hidden` content wrapper is ~4px too short.

| | |
|---|---|
| **Component** | `@nysds/nys-accordion` (`nys-accordionitem`) |
| **Version** | 1.19.0 |
| **Severity** | Visual / minor — bottom border + last line of content clipped |
| **Repro** | Bordered accordion with an item set to `expanded` on initial render |

## Steps to Reproduce

```html
<nys-accordion bordered>
  <nys-accordionitem heading="What is changing?" expanded>
    <p>We are testing USWDS primitives under the NYSDS token API.</p>
  </nys-accordionitem>
  <nys-accordionitem heading="Does it affect components?">…</nys-accordionitem>
</nys-accordion>
```

**Expected:** The expanded panel shows its full content with the bottom border fully visible.
**Actual:** The bottom border (and a sliver of content) is cut off. Toggling the item closed and re-opening it fixes the height, because by then the border is present when `_updateHeight()` runs.

## Root Cause

The bug is a **timing race between height measurement and border application**, compounded by the fact that height is never recalculated when `bordered` changes.

**1. The border lives on a child of the measured/clipped element.**

`_updateHeight()` measures and pins the height on the **outer** `.nys-accordionitem__content` (which has `overflow: hidden`):

`packages/nys-accordion/src/nys-accordionitem.ts:118`
```ts
private _updateHeight() {
  if (!this._contentContainer) return;
  if (this.expanded) {
    const slotHeight = this._contentContainer.scrollHeight;   // outer .nys-accordionitem__content
    this._contentContainer.style.height = `${slotHeight}px`;
  } else { ... }
}
```

But the border is applied to the **inner** `.nys-accordionitem__content-slot-container`:

`packages/nys-accordion/src/nys-accordion.scss:137`
```scss
:host([bordered]) .nys-accordionitem__content-slot-container {
  border: var(--_nys-accordion-border-width) solid var(--_nys-accordion-border-color);  /* 2px top + 2px bottom */
}
```

**2. `bordered` is applied by the parent *after* the child has already measured.**

The child item measures its height during its own first render:
- `firstUpdated()` adds a `slotchange` listener → `_updateHeight()` (`nys-accordionitem.ts:62-76`)
- `updated()` runs `_updateHeight()` because `expanded` is in `changedProperties` (`nys-accordionitem.ts:78-82`)

At that point `bordered` is still `false` on the item — the parent `nys-accordion` only pushes it down later, in its own update cycle:

`packages/nys-accordion/src/nys-accordion.ts:71-75, 112-116`
```ts
updated(changedProperties) {
  if (changedProperties.has("bordered")) this._applyBordered();
}
private _applyBordered() {
  this._getAccordionItems().forEach((item) => { item.bordered = this.bordered; });
}
```

So the height is locked to the **pre-border** `scrollHeight`. When the border subsequently appears (+4px), the outer container's `overflow: hidden` clips the extra height — exactly the cut-off bottom in the screenshot.

**3. Height is never recalculated when `bordered` changes.**

The item's `updated()` only recalculates on `expanded` changes, not on `bordered`:

`packages/nys-accordion/src/nys-accordionitem.ts:78-82`
```ts
updated(changedProperties) {
  if (changedProperties.has("expanded")) {   // ⚠️ does not also watch "bordered"
    this._updateHeight();
  }
}
```

This is why a manual close→open cycle "fixes" it: the second measurement happens after the border is present.

## Suggested Fix

Any of these (ideally the first + one of the others):

1. **Recalculate height when `bordered` changes** in `NysAccordionItem.updated()`:
   ```ts
   updated(changedProperties) {
     if (changedProperties.has("expanded") || changedProperties.has("bordered")) {
       this._updateHeight();
     }
   }
   ```
2. **Measure the element that actually carries the border.** Use the inner `.nys-accordionitem__content-slot-container`'s `offsetHeight` (which includes its own border box) instead of the outer wrapper's `scrollHeight`, so border width is always included regardless of timing.
3. **Apply `bordered` before first measurement** — e.g. have the item read `bordered` from its parent in `connectedCallback()` / set it reflectively before the initial `_updateHeight()`, so the first measurement already includes the border.

## Notes

- A `ResizeObserver` on the content container would make the pinned height self-healing against this entire class of late-layout changes (fonts loading, border/padding changes, responsive reflow) and is worth considering for a more robust fix.

NovaCat35:
# Summary

Fix issue with height not accounting for `bordered` and `expand` height, as well as new content additions that are programmatically generated.

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1633 🎟️

### DIFF
--- a/.github/workflows/deploy-github-pages.yaml
+++ b/.github/workflows/deploy-github-pages.yaml
@@ -26,7 +26,7 @@ jobs:
       # Set up Node
       - uses: actions/setup-node@v6
         with:
-          node-version: '24.15.0'
+          node-version: '24.17.0'
 
       #👇 Add Storybook build and deploy to GitHub Pages as a step in the workflow
       - uses: bitovi/github-actions-storybook-to-github-pages@v1.0.3
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -4576,7 +4576,7 @@
               "type": {
                 "text": "\"md\" | \"lg\" | \"full\""
               },
-              "default": "\"md\"",
+              "default": "\"full\"",
               "description": "Input width: `md` (200px), `lg` (384px), `full` (100%).",
               "attribute": "width",
               "reflects": true
@@ -5113,7 +5113,7 @@
               "type": {
                 "text": "\"md\" | \"lg\" | \"full\""
               },
-              "default": "\"md\"",
+              "default": "\"full\"",
               "description": "Input width: `md` (200px), `lg` (384px), `full` (100%).",
               "fieldName": "width",
               "propName": "width"
@@ -7112,6 +7112,16 @@
             },
             {
               "kind": "field",
+              "name": "agencySubheading",
+              "type": {
+                "text": "string"
+              },
+              "default": "\"\"",
+              "description": "Optional subheading displayed below the agency name.",
+              "attribute": "agencySubheading"
+            },
+            {
+              "kind": "field",
               "name": "homepageLink",
               "type": {
                 "text": "string"
@@ -7148,6 +7158,16 @@
               "propName": "agencyname"
             },
             {
+              "name": "agencySubheading",
+              "type": {
+                "text": "string"
+              },
+              "default": "\"\"",
+              "description": "Optional subheading displayed below the agency name.",
+              "fieldName": "agencySubheading",
+              "propName": "agencysubheading"
+            },
+            {
               "name": "homepageLink",
               "type": {
                 "text": "string"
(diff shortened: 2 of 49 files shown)


esteinborn (approved):
LGTM;

PR #1652, opened 2026-06-25
Bug/translate icon missing

### LINKED ISSUE #1651: Bug: Translate Icon missing

emilygorelik:
Fix missing translate icon appearing on circle button due to logic of how slotted icons on button work. a previous fix updated the prefix and suffix icons for slotted buttons but the fix was missing from circle icon. this fixes that.
image

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -2229,6 +2229,15 @@
             },
             {
               "kind": "field",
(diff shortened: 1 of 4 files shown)


esteinborn (approved):
LGTM;

PR #1682, opened 2026-07-02
docs(nys-globalheader): clarify when to enable `nysLogo`

plasticmind:
## Summary

The `nys-globalheader` `nysLogo` prop was documented only as *"Toggles the NYS brand mark"* — with no guidance on **when** to use it. Because the prop name actively invites enabling it on "a NYS site," the brand mark was getting turned on in standard page layouts where `nys-unavheader` already provides NYS branding one row above. This surfaced repeatedly through AI-assisted authoring (the MCP server re-serves this description verbatim).

The default was already correct (`false`); the gap was purely documentation.

## What changed

Clarified in three places that flow from a single source (the JSDoc):

- **`nys-globalheader.ts`** — expanded the `nysLogo` JSDoc with intent + default.
- **`nys-globalheader.mdx`** — added a note to the *NYS Brand Mark* section and Do/Don't lists.
- **`custom-elements.json`** + React type wrappers — regenerated via `npm run cem` so the manifest (and the MCP server that reads it) carry the new guidance.

## The rule

`nysLogo` is **off by default**. Enable it **only for internal, state-employee (back-office) apps that omit `nys-unavheader`**. Any resident-facing app — even one requiring login — should keep `nys-unavheader` for trust and leave `nysLogo` off.

## Notes

- Docs-only; no behavior or API change.
- The reference site (`nysds-site`) already documents this correctly, so no companion PR is needed there.

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -7330,7 +7330,7 @@
                 "text": "boolean"
               },
               "default": "false",
-              "description": "Toggles the NYS brand mark",
+              "description": "Displays the NYS brand mark in the header. Off by default.\n\nEnable only for internal, state-employee (back-office) applications that omit\n`nys-unavheader`. Any resident-facing app — even one requiring login — should\nkeep `nys-unavheader` for trust and leave this off.",
               "attribute": "nysLogo"
             },
             {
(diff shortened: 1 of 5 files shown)


esteinborn (approved):
LGTM;

PR #1747, opened 2026-07-09
Fix: `nys-datepicker` to dispatch input event on "clear" & "today" button

### LINKED ISSUE #1653: FIX: Datepicker -  Today and Clear buttons don't fire an update event
## Problem Description:

When using the Date picker component on a page that would compose a POST Request, versus making a multipart form submission, we noticed that clearing the input from the Calendar Widget does not trigger an additional nys-input event. This makes it difficult to consume outside a form reliably. This is also true for the Today button.

Our team primarily works in Vue, so for some controls or components we create submissions from users using Axios POST requests instead of an HTML form. It would be useful if we can reliably clear our reactive reference to the Date object if the user clicks "Clear" from the Calendar Widget.

## Recreating the issue:

When creating the writeup for the NysIcon library vite issues we encountered, I had created a public repository for demoing our usage. I have enabled github pages and a simple workflow to deploy the main branch on push.

This is available at, https://mlaufer-nysed.github.io/mlaufer-nysed-nysds-sample/

This simply sets up a page with a Datepicker, and a button to `window.alert()` our reactive ref to the Datepicker's model value. There is also a counting ref, displaying how many times the `nys-alert` event has been captured on this page.

Script:
```ts
// event counter
const eventsFired = ref(0);
// model value from <nys-datepicker>
const dateValue: Ref<Date | undefined> = ref(undefined)
// function to respond to event
const onInputEvent = (e: CustomEvent) => {
  const { id, value } = e.detail
  dateValue.value = value
  eventsFired.value++
}
// button on click event
const windowAlert = (s: string) => {
  window.alert(s)
}
```
Template:
```vue
<template>
  <div>
    <p>Events Fired: {{ eventsFired }}</p>
    <nys-datepicker @nys-input="onInputEvent" />
    <nys-button @click="windowAlert(`Model Value: ${dateValue}`)" label="Alert Model Value" />
  </div>
</template>
```

You can see from interacting with the Datepicker, when;
1. selecting a new date either via the calendar
2. Typing a new date into the text input box 

the Events fired counter increments, and you will be able to log the selected value. Selecting Today, or Clear, does not.

Attempting to log the model value after clearing the Datepicker, holds the previous value still.

---

Our current workaround was one of two solutions:

1. Wrapping the <nys-datepicker> WebComponent in a vue component that disallows the Today / Clear buttons from within the Calendar Widget, and instead adds a clear button externally to set the value via the attribute / prop.
```ts
const props = withDefaults(defineProps<NysDatePickerPropsInterface>(), {
  hideTodayButton: true;
  hideClearButton: true;
});
```
2. Utilizing nys-input event like normal, but also hooking into the blur event, for purpose of capturing when the user opens the Calendar Widget to select clear / today, and grabbing the value from the HTMLInputElement
```ts
// handles catching clearing or setting via today
const onBlur = (e: Event) => {
  const target = e.target as HTMLInputElement;
  // 'cast' from string to date
  model.value = target.value as unknown as Date;
};
```

We opted for the second option as we didn't want to disallow the native clear button all the time. However, both felt to me a tad on the hacky side. Any recommended usages or updating the Datepicker to emit those events on the Today/Clear button clicks as well would be much appreciated.

We appreciate your help and support!

Thanks,
Michael

NovaCat35:
# Summary
> Thank you [@mlaufer-nysed](https://github.com/mlaufer-nysed) for this contribution.

This PR fixes an issue with `nys-datepicker`'s "Today" and "Clear" buttons not firing an nys-input update event.

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1653 🎟️

## Testing
Done on the [React Demo](https://github.com/ITS-HCD/nysds-react-demo) 

## Screenshot
A quick demo showing how I try to test the event on the inspector and making sure the event dispatches for the two buttons

Screenshot 2026-07-09 at 10 02 56 AM


packages/nys-datepicker/src/nys-datepicker.ts
```diff
@@ -710,9 +710,9 @@ export class NysDatepicker extends LitElement {
     if (this.disabled) return;
 
     this._setTodayDate();
-    if (this._hasUserInteracted) {
-      this._validate();
-    }
+    this._hasUserInteracted = true;
+    this._validate();
+    this._dispatchInputEvent();
```
emilygorelik, line 715: should these be custom events?

NovaCat35: ~~Great point, good callout there~~
After some consideration, this button's main purpose is to change the value of the input. The correct dispatch event should be the `nys-input.`

emilygorelik (dismissed):
LGTM

plasticmind (changes requested):
The fix is right. One small fix, an explanation, and a test to add.

### The fix is solid ✅

Today and Clear now mark the field interacted, validate, and fire `nys-input` — same as `_handleDateChange` and `_handleInputChange`. Two things I checked:

- **No double-firing.** `_setValue` (via `_setTodayDate`) doesn't dispatch `nys-input`, so your one `_dispatchInputEvent()` is the only one.
- **The test flip is correct.** Clear-on-required going `showError: false` → `true` is right — once Clear counts as an interaction, eager validation should show the required error.

### 🔴 Fix before merge: the generated files drifted

Four of the six files — `custom-elements.json`, `NysDatepicker.d.ts`, `NysDatepicker.js`, `nysds-jsx.d.ts` — are generated from the component source, not hand-edited.

They advertise two events, `nys-datepicker-today` and `nys-datepicker-clear`, plus two methods, that don't exist in the component. Your source only fires `nys-input`; I checked the branch head. So `NysDatepicker.js` wires up:

```js
useEventListener(ref, "nys-datepicker-today", props.onNysDatepickerToday);
useEventListener(ref, "nys-datepicker-clear", props.onNysDatepickerClear);
```

A React dev wouldn't get an error, just no event. It seems like you didn't regenerate after you switched to `nys-input`.

Fix: regenerate on a clean tree (`npm run cem` + the React wrapper gen, both in `build:all`). That drops the four phantom entries. Don't hand-edit generated files — fix the source and re-run.

### 🟡 Add: a test that the event fires

`handles Today button click` / `handles Clear button click` only check `el.value` — they'd pass even if dispatch broke. The file already has the pattern (`oneEvent`, lines 88 and 352):

```ts
const eventPromise = oneEvent(el, "nys-input");
todayBtn.dispatchEvent(new MouseEvent("click", { bubbles: true }));
const event = await eventPromise;
expect(event.detail.value).to.exist;   // undefined for Clear
```

One for each button.

### Bottom line

1. Regenerate the four files so the phantom events disappear.
2. Add the two dispatch tests.

esteinborn (approved):
LGTM;

PR #1750, opened 2026-07-10
Fix: Vertical Grid Gap - use native rowgap

### LINKED ISSUE #1720: BUG: grid-gap classes applying negative V margin when none existed before
This issue currently shows up on the reference website. you can see it in the /about/updates page header and the /about/updates/all headers.

The root cause is coming from the new changes to grid-gap classes where they are adding negative margin to the Y axis, causing the "main" container to pull upwards -1rem. causing the visual bug that shows the section header as "shorter." It's not shorter, but is instead being overwritten in -1rem as white BG (the default BG color for the Main element).

Please look at how negative margins in the Y axis work and ensure that we're not pulling negative margins on containing elements.

Image

emilygorelik:
Update vertical logic to use native gridgap to avoid overriding the existing margins and paddings that can be set on containers


plasticmind (approved):
LGTM. 🚀 Reproduced the bleed and the fix.

Negative top/bottom margin bled outside the flex container and pulled the row up over whatever sat above it. `row-gap` spaces the wrapped lines with no bleed, and leaves column widths alone (unlike `column-gap`).

Spacing comes out identical — `$half = $value/2`, so `-$half` margin + `$half` padding becomes `row-gap: $value` at every tier. Built it both ways: `row-gap` everywhere, no `margin-top` left on any `grid-gap-y` rule. Repro'd it too — develop overlaps the header, branch sits flush with a clean gap. Matches your before/after. 👍🏻

🟡 Known limit, not a blocker: horizontal still uses the negative left/right margin, so a row's own left/right margin still gets overridden. Vertical axis only here.

Ship it. 💪🏻

PR #1751, opened 2026-07-10
Enhancement: Tabpanel - LightDOM

### LINKED ISSUE #1569: Tab: components in tab-panel css and js not working

### LINKED ISSUE #1714: LightDOM: tabpanel

emilygorelik:
image
image


plasticmind (approved):
LGTM. 🚀 Same light-DOM move as #1761 (the table), applied to tabs, which lets a form inside a panel be reached and driven by page JS.

### 👍🏻 Good stuff

- **The slot cascade settles — no infinite loop.** Assigning `slot="tab"`/`"panel"` fires more `slotchange` events, but `_syncChildren` only calls `setAttribute` when the slot is wrong (`nys-tabgroup.ts` `_syncChildren`), so once the slots are right, nothing mutates and it stops. The guard is what makes this safe — good that it's there.
- **Selection survives a re-sync.** `_applySelection` writes the `selected` attribute on the active tab (`:398`), and `_syncChildren` re-reads it to pick the index. So adding a tab/panel later via JS won't snap selection back to the first tab.
- **Pairing is stable when run repeatedly (idempotent).** First pass pairs by explicit `aria-labelledby`, then fills the rest by source order; `_applySelection` then stamps `aria-labelledby` on every panel, so the next run reads those and reproduces the same pairing. `_pairPanels` is clean and well-commented.
- **Panels aren't physically reordered anymore** — only the paired array drives which one gets `hidden`, and one shows at a time, so the visual result matches with less DOM churn.
- **The tricky `[hidden]` bit is handled right.** `display` stays in the shadow sheet (`:host([hidden]) { display: none }`) and the document-adopted light sheet deliberately leaves `display` alone — otherwise an author-level rule would beat the UA `[hidden]` rule and inactive panels wouldn't hide. Tricky part of styling from a document sheet, but your comment calls it out.
- **Tests are strong:** light-DOM projection, shadow stays empty, consumer CSS/JS reaching panel content, a real form with a change-driven rule, and non-tab children left in the default slot.

### 🟡 A few small notes

1. During the initial mount, `_syncChildren` → `_applySelection` runs a few times (each child leaving the default slot and landing in a named slot fires `slotchange`). It's idempotent so it's harmless — if you want fewer redundant passes you could bail when nothing actually changed. Perf only.
2. Panel styling is now document-level and overridable (that's the feature) — worth a changelog note about the encapsulation change, same as #1761.
3. Known limitation we may want to address in the future: the light sheet reads `--_nys-tabpanel-background-color`, which isn't declared on the host, so panels are transparent unless a consumer sets it. Not new here — just surfaced now that the "Custom Styling" story leans on it. If a default panel background was ever intended, give it a fallback.

PR #1753, opened 2026-07-10
Fix: Checkbox setError resolves on selection

emilygorelik:
Before:

image

After:

image

image


plasticmind:
LGTM. 🚀 Reproduced both sides.

The bug: a group that isn't `required` but has `showError` set never cleared after a selection. `_manageRequire()` hit `if (!this.required) return` and bailed before touching validity. Now the non-required path clears validity and drops `showError` — same as what `_checkOtherInputs` already does at `nys-checkboxgroup.ts:578`.

Checked it: non-required group with `showError`, check a box, expect the error gone. Fails on develop, passes here. Full checkbox suite still green.

Pushed a regression test for it so the path stays covered — `nys-checkbox.test.ts`, "clears showError when a checkbox is selected (not required)". 35/35 on the branch.

plasticmind (approved):

PR #1755, opened 2026-07-11
fix: repair broken CSS custom property references

plasticmind:
## Summary

A CSS audit of the built stylesheet flagged several `var()` references to custom properties that don't exist. These silently fell back to hardcoded values (or did nothing), so they ignored token overrides and theming.

### Fixes

- **Stale `--nys-type-*` names → `--nys-font-*`** in `nys-datepicker`, `nys-tooltip`, `nys-accordion`, `nys-globalheader`, and `nys-unavheader` (pre-rename token names; components were stuck on their px fallbacks). Also fixed two adjacent `--nys-typography-*` lineheight/letterspacing names in `nys-unavheader`.
- **Heading font-weight utilities**: `.nys-font-h1`–`h6` referenced `--nys-font-weight-h1`–`h6`, which never existed — always rendered the `bold` fallback. Now uses `--nys-font-weight-bold`. No visual change.
- **h1 letter-spacing**: the heading map in `reset.scss` referenced `--nys-font-letterspacing-negative-300`, which doesn't exist (scale stops at `negative-200`), so h1 got **no** letter-spacing. Now uses `negative-200`, matching h2/h3. ⚠️ Visual change: h1 tracking tightens slightly.
- **List markers**: `li::marker` referenced `--nys-color-theme-primary` (doesn't exist; theme scale is `faint/weaker/weak/mid/default/strong/stronger`), so markers rendered the black fallback. Now uses `--nys-color-theme`. ⚠️ Visual change: markers become theme-colored, which appears to be the original intent.
- **Token docs**: the transparency checkerboard in `Design-Tokens.mdx` used `--nys-color-neutral-40`; the neutral scale has no 40 step. Now uses `--nys-color-neutral-10` (matches the existing `#f6f6f6` fallback — that's neutral-10, not 50).

### Testing

- Repo-wide grep confirms no remaining references to the nonexistent variables
- Recommend a visual pass in Storybook on headings and lists, since the last two fixes intentionally change rendering


packages/styles/src/core/reset.scss
```diff
@@ -219,7 +219,7 @@ li {
 }
 
 li::marker {
-  color: var(--nys-color-theme-primary, #000000);
+  color: var(--nys-color-theme, #000000);
```
emilygorelik: theme fallback should be our blue, not black. Did you mean ink or to set fallback to `#154973`

src/stories/Design-Tokens.mdx
```diff
@@ -819,10 +819,10 @@ export const hiddentokens = [
         }
         .checkerboard {
           background-image:
-            linear-gradient(45deg, var(--nys-color-neutral-40, #f6f6f6) 25%, transparent 25%),
-            linear-gradient(-45deg, var(--nys-color-neutral-40, #f6f6f6) 25%, transparent 25%),
-            linear-gradient(45deg, transparent 75%, var(--nys-color-neutral-40, #f6f6f6) 75%),
-            linear-gradient(-45deg, transparent 75%, var(--nys-color-neutral-40, #f6f6f6) 75%);
+            linear-gradient(45deg, var(--nys-color-neutral-50, #f6f6f6) 25%, transparent 25%),
```
emilygorelik: `f6f6f6` is the raw value for `--nys-color-neutral-10` not `50`

emilygorelik (changes requested):
some fallback values are out of alignment. see comments

packages/styles/src/core/reset.scss
plasticmind: Good catch — `--nys-color-theme` resolves to `#154973`, so that's the fallback now. Black was doing it dirty. 👍🏻

src/stories/Design-Tokens.mdx
plasticmind: Yep — `#f6f6f6` is neutral-10, `50` is `#ededed`. Pointed the var at `--nys-color-neutral-10` so it matches the fallback and the checkerboard stays the same shade. 🎯

plasticmind:
Both fallbacks fixed and pushed. Also reviewed the rest of the fallbacks against the real token values... everything looks good. Ready for another look when you get a sec. 🙏🏻

emilygorelik (approved):
LGTM

PR #1758, opened 2026-07-13
fix ids and classes not pulled into basic demo

### LINKED ISSUE #1759: Fix: basic demo classes and ids not populating

emilygorelik:
the class and id props were not be default getting pulled into the basic demo of storybook. This fix to the generation script fixes that issue


plasticmind:
LGTM. 🚀 Ran it end to end.

The fix: the interactive Basic story rebuilt the root tag from prop bindings only and threw away everything else — so `class` and `id` from the example never reached the rendered demo. That's what broke modal's wiring, where the open button does `onclick="...querySelector('.modal1')"` and the selector had nothing to grab. `generate-stories.mjs:340` now keeps the static attrs and prepends them before the bindings — same thing the child-tag path already does a few lines down.

Checked by running `npm run gen-stories` on the branch → clean tree, zero drift. So the story changes are exactly what the generator emits, and every id/class traces back to the `@example` JSDoc in the source (`nys-modal.ts:25`, `nys-datepicker.ts:29`). No regen debt. 👍🏻

Folded in one more fix while here: a few props were silently missing from the Basic controls. `parseStringUnionOptions` splits the type on `|`, so a union written with a leading pipe — `| "filled" | "outline" …`, how several of our components format it — starts with an empty segment that fails the quote match, and the whole prop gets dropped. `.filter(Boolean)` clears that empty segment. Restores selects for button's `variant`/`type`/`target`, alert's `type`, badge's `intent`, and `width` on select/textarea/textinput. Button's `variant` is a real control now instead of the hardcoded `variant="filled"`.

PR #1761, opened 2026-07-14
Enhancement: Table - LightDOM

### LINKED ISSUE #1594: Potential issue: nys-toggle events within nys-table
Possible nysds issue.  It appears that events attached to toggles that are inside the nys-table do not get fired.   I didn't try it for other elements.

Simple example attached.

[nysdstoggle.txt](https://github.com/user-attachments/files/28601902/nysdstoggle.txt)

### LINKED ISSUE #1650: BUG: `nys-table` renders a shadow-DOM clone, so edits to interactive cells are lost (forms submit stale values)
## Summary

`nys-table` deep-clones the slotted `<table>` into its shadow DOM and displays the clone, leaving the author's original table hidden in the light DOM. When a cell contains an interactive control, the user edits the **clone** while a surrounding `<form>` and any script read the **hidden original** — so user input is silently dropped on submit.

This is the **root cause** behind a family of `nys-table` symptoms already reported:
- #1594 — toggle events inside `nys-table` don't fire (listeners live on the hidden original, not the displayed clone).
- #1490 — framework (preact/react) updates don't reflect in the table (the framework reconciles the original; the clone is a dead copy).

This ticket captures the underlying mechanism and a verified form-state reproduction.

## Component / Area

`nys-table`

## Current behavior

- The slotted table is kept in the light DOM but hidden via `<slot style="display:none">` — `packages/nys-table/src/nys-table.ts:416`.
- On slot change / mutation the component deep-clones it — `cloneNode(true)` (`nys-table.ts:97`) — and renders the clone in a shadow-DOM container, rebuilt from scratch each time via `container.innerHTML = ""` (`nys-table.ts:88`, `:99-105`).
- The visible inputs therefore live in the shadow root; the form-associated originals live (hidden) in the light DOM.

## Verified reproduction

Table with an `<input>` in a cell, inside a `<form>`, using `@nysds/nys-table@1.19.0`. User edits the visible Qty field to `999`:

```
user sees/edits (shadow clone): 999
hidden light-DOM original     : 1
what the form SUBMITS (qty)   : 1
```

The form submits `1` — the value the user never sees — because the edited input is inside a shadow root and isn't associated with the light-DOM form. (Confirmed in a browser, not inferred.)

Minimal markup:

```html
<form id="myform">
  <nys-table sortable>
    <table>
      <caption>Order</caption>
      <thead><tr><th>Item</th><th>Qty</th></tr></thead>
      <tbody>
        <tr><td>Widget</td><td><label for="qty">Qty</label><input id="qty" name="qty" type="number" value="1"></td></tr>
      </tbody>
    </table>
  </nys-table>
</form>
```

## Expected behavior

The table the author provides is the table the user interacts with. Edits to form controls in cells are reflected in the light-DOM original and submitted with a surrounding form.

## Same root cause, related impacts

Because the visible table is a fresh `cloneNode` rather than the author's element:
- Event listeners attached to cell content don't transfer to the clone — `cloneNode` never copies listeners (this is #1594).
- Framework-rendered tables (React/Vue/Angular/preact) reconcile and delegate events against the light-DOM original, not the shadow clone (this is #1490).
- Transient state (focus, selection, scroll) is wiped on every re-clone (`nys-table.ts:88`).

> Note: an earlier hypothesis that this causes duplicate-`id`/`<label for>` breakage was tested in a browser and **ruled out** — shadow-DOM id scoping keeps in-cell `<label for>` working. Not an issue.

## Proposed direction

Render and enhance the table **in place in the light DOM** instead of cloning into shadow DOM:

- Light-DOM render root (or keep shadow only for the download-button chrome).
- Sort by reordering the real `<tr>` nodes; guard the `MutationObserver` (disconnect during self-mutation) to avoid re-entrancy.
- Move styles from `static styles` (shadow-only) into a namespaced light-DOM stylesheet (e.g. `@nysds/styles` scoped under `.nys-table`).

**Tradeoff:** loses automatic shadow-DOM style encapsulation (mitigate via `.nys-table` namespacing + design tokens) and adds observer re-entrancy handling — in exchange for correct form/interactivity behavior and consumer-styleable cells. `::slotted()` is **not** a viable middle path: it can't style descendants (`<td>`/`<th>`) of a slotted table.

## Related issues

- Same root cause as #1594 (toggle events not firing) and #1490 (framework updates not reflected).

## Environment

`@nysds/nys-table@1.19.0`; framework-agnostic, most severe inside a `<form>` or under a JS framework.

### LINKED ISSUE #1713: LightDOM: table

### LINKED ISSUE #1719: ShadowDOM: table (Cloning Removal)
## Summary
In our effort to align with the native HTML implementation and avoid cutting user/dev off from controlling what they put into our “slots” for our `nys-`components, we hereby will remove all slots cloning implementation going forward.

emilygorelik:
Update table component to lightDOM to allow JS and CSS to work inside.


plasticmind (approved):
**LGTM. 🚀** The table stays in the light DOM and is enhanced in place instead of cloned into the shadow DOM. That fixes the bug where cloned embedded components (badges, buttons) lost their listeners.

### 👍🏻 Good suff:

- **Safe to run more than once (idempotent).** The old code cloned a fresh table each time, so re-running enhancement didn't matter. Now it edits the *real* table in place, so enhancement has to be safe to repeat — and it is: `_normalizeTableDOM` returns early when `thead && tbody` already exist (`nys-table.ts:233`), and `_addSortIcons` skips a `th` that already has a sort-button (`:307`). So the three things that trigger it on mount (firstUpdated + slotchange + observer) can't double up the structure or icons.
- **Pauses the watcher during its own edits (re-entrancy guard).** When the code edits the table, it first disconnects the MutationObserver, does the edit, then reconnects (`_withObserverPaused`, with `finally` so it reconnects even on error). Without that, the observer would see our own edits and re-trigger enhancement in a loop. `_enhancing` guards nested calls.
- **No leaks, safe to server-render (SSR).** `disconnectedCallback` disconnects the observer. Light styles adopt once via `_lightSheet`, skipped when `document` is undefined.
- **Styling resolves.** Selectors are scoped to the `nys-table` tag; the `--_nys-table-*` variables on `:host` inherit down into the slotted table.
- **Generated files match source.** No drift. 💪🏻 
- **Tests were updated** to target `el.querySelector` and use one non-cloned copy with a working listener.

### 🟡 A few small notes

1. **Please add a test for changing the table after mount** — mutate it once it's mounted and assert no duplicate sort-buttons. In-place enhancement is the core of this refactor and it's the one new path that isn't covered.
2. Cell styles are now document-level. Consumers can reach `nys-table td/th` — intended — but NYSDS's rules now compete by specificity instead of being sealed in the shadow DOM. Note it in the changelog.
3. Known limitations we may want to addres in the future: if users add a new row via JS, they won't get the sorted-column highlight while a sort is active.
4. `_handleSlotChange` runs 2–3× on mount. It's harmless, but you can drop the explicit `firstUpdated` call for better performance.

PR #1762, opened 2026-07-14
Feature/iconlist

### LINKED ISSUE #1766: COMPONENT: Icon List

emilygorelik:
> This should be released in the same version as nys-card

### Basic
image

### Divider
image

### Secondary Label
image

### Checklist
image


plasticmind (commented):
Super close. One change and a couple small notes:

## 🟡 Things to change:

**`_handleSlotChange` deletes children it doesn't recognize.** Any direct child that isn't a `<nys-iconlistitem>` gets `el.remove()`d (`nys-iconlist.ts:~88`). Two problems with deleting rather than ignoring:
1. If someone drops in a wrapper, a heading, or a stray element, it silently vanishes. That's a confusing debugging session with no error to guide you.
2. If a framework like React owns a node in that light DOM and the component removes it, React's next reconcile tries to touch a node that's gone and throws `NotFoundError`

`assignedElements` already skips text nodes so whitespace is safe, just fix by not deleting. Warn and leave, or hide with `::slotted(:not(nys-iconlistitem))` in CSS.

## Small notes

- **Standalone item is an orphan `listitem`.** `nys-iconlistitem` sets `role="listitem"` in `connectedCallback` regardless of parent, so using one outside a `nys-iconlist` is invalid ARIA (a listitem with no list). Intended use is inside the list, so this is minor — worth a line in the docs, or gate the role on having a list parent.
- **`console.warn` ships to production.** Fine as a dev nudge, but paired with the deletion above it's doing real work silently. If you keep the warn, drop the delete.

## Bottom line

1. Stop deleting unrecognized children — warn or hide, don't `remove()` 🟡
2. Doc (or guard) the standalone-item role

mendeza:
Yes, [it worked](9a33af68e9e844bf910e6e13555ad4abef194cdf). Bravo!

packages/nys-iconlist/src/nys-iconlistitem.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  private _handleSecondarySlotChange(e: Event) {
+    const slot = e.target as HTMLSlotElement;
+    const hasSecondary = slot.assignedNodes({ flatten: true }).length > 0;
+    this.toggleAttribute("data-has-secondary", hasSecondary);
+  }
+
+  render() {
+    return html`
+      <div class="nys-iconlistitem">
```
NovaCat35, line 55: Ah okay I see. Given Aaron is expressing his approval, just want to double-check that going with `div` over an actual HTML semantic list (`ul`, `ol`, `li`) is the right approach?

packages/nys-iconlist/src/nys-iconlist.ts
```diff
@@ -0,0 +1,138 @@
...
+import { property } from "lit/decorators.js";
+import "./nys-iconlistitem";
+// @ts-ignore: SCSS module imported via bundler as inline
+import styles from "./nys-iconlist.scss?inline";
+
+let componentIdCounter = 0;
+
+// The list renders in light DOM, so its styles are adopted into whichever
+// root (document or containing shadow root) the element lives in — once per root.
+const lightSheet = new CSSStyleSheet();
```
NovaCat35: Is line 11 to 13 and the const `styledRoots` used to implement styling in the lightDOM?  

I'm not sure if it's implementing similar code like
``` js
let _lightSheet: CSSStyleSheet | null = null;
// Injects the lightDOM styling for the scss for
// styling CSS into the adopted/constructed stylesheet.
function adoptLightStyles() {
  if (_lightSheet || typeof document === "undefined") return;
  _lightSheet = new CSSStyleSheet();
  _lightSheet.replaceSync(lightStyles);
  document.adoptedStyleSheets = [...document.adoptedStyleSheets, _lightSheet];
}
```
Just want an explanation of this part of the code

packages/nys-iconlist/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@nysds/nys-iconlist",
+  "version": "1.19.2",
```
NovaCat35: Update version

packages/nys-iconlist/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "scripts": {
+    "dev": "tsc --emitDeclarationOnly && vite",
+    "build": "vite build",
+    "test": "vite build && wtr",
+    "build:watch": "tsc --emitDeclarationOnly && vite build --watch",
+    "test:watch": "vite build && wtr --watch",
+    "lit-analyze": "lit-analyzer 'src/**/*.ts'"
+  },
+  "dependencies": {
+    "@nysds/nys-icon": "^1.19.2"
```
NovaCat35: update version

packages/nys-iconlist/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "vite": "^7.3.1"
+  },
+  "keywords": [
+    "new-york-state",
+    "design-system",
+    "web-components",
+    "lit",
+    "nys",
+    "iconlist",
+    "forms"
```
NovaCat35: Not a form component

packages/nys-iconlist/src/nys-iconlist.ts
```diff
@@ -0,0 +1,138 @@
+import { LitElement, PropertyValues } from "lit";
+import { property } from "lit/decorators.js";
+import "./nys-iconlistitem";
+// @ts-ignore: SCSS module imported via bundler as inline
+import styles from "./nys-iconlist.scss?inline";
+
+let componentIdCounter = 0;
```
NovaCat35, line 7: Name it from `componentIdCounter` to a more specific name like `iconListIdCounter`

packages/nys-iconlist/src/nys-iconlistitem.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  /* Without this the host defaults to display:inline while its shadow content
+     is block-level, which splits it into anonymous blocks and drops its
+     accessibility node — taking role="listitem" with it. */
+  display: block;
+
+  color: var(--nys-color-text);
+
+  /* UI/Large/Semibold */
+  --_nys-iconlistitem-font-family: var(--nys-font-family-ui, "Proxima Nova");
```
NovaCat35: Give more fallbacks:
```css
var(
    --nys-font-family-ui,
    var(
      --nys-font-family-sans,
      "Proxima Nova",
      "Helvetica Neue",
      "Helvetica",
      "Arial",
      sans-serif
    )
  );
```

packages/nys-iconlist/src/nys-iconlistitem.scss
```diff
@@ -0,0 +1,65 @@
...
+
+:host([divider]) .nys-iconlistitem {
+  padding-bottom: var(--_nys-iconlistitem-spacing--divider);
+  margin-bottom: var(--_nys-iconlistitem-spacing--divider);
+  border-bottom: var(--_nys-iconlistitem-width--divider) solid
+    var(--_nys-iconlistitem-color--divider);
+}
+
+.nys-iconlistitem__icon {
+  padding: var(--nys-space-50, 4px);
```
NovaCat35: Make this a private `var`

packages/nys-iconlist/src/nys-iconlist.figma.ts
```diff
@@ -0,0 +1,8 @@
+import figma, { html } from "@figma/code-connect/html";
+
+figma.connect("<FIGMA_ICONLIST>", {
+  props: {
+    // TODO: Add props here
```
NovaCat35, line 5: Please add some boilerplate code connect (again, code connect doesn't need to be completely connected, just needs to be shown as an example on Figma).

NovaCat35 (changes requested):
First review pass

packages/nys-iconlist/src/nys-iconlistitem.ts
NovaCat35: @mendeza Any thoughts on this?

packages/nys-iconlist/src/nys-iconlist.ts
esteinborn: Ditto to @NovaCat35 comments. We created a pattern for loading stylesheets into the lightDOM, this is not that approach.

esteinborn:
What is in light and what is in shadowDOM?

esteinborn (changes requested):
see comments

esteinborn (changes requested):
Looking at the lightDOM implementation, you are loading from the same file as previously: `nys-iconlist.scss`, but any lightDOM should be defined in `nys-iconlist.light.scss`

Please check these filenames, and separate out what should be in lightDOM and what can stick in shadowDOM.

esteinborn (approved):
LGTM;

Issue #1721, opened 2026-07-06
FIX: nys-radiogroup 1.19 rework broke radios in tables & custom layouts (agreed fix inside)


plasticmind:
> **Status:** Blocked by #1632 (ElementInternals a11y layer). This fix should be built on top of #1632's new base classes and its `associateHost` helper, so it should be picked up **after #1632 merges** — see the implementation notes at the bottom.

## What's happening

Starting in version **1.19.0**, radio buttons stopped working when they're placed inside a table or any custom layout. A team upgrading from 1.18.3 to 1.19.2 ran into this: they have a table with a radio button in each row so a user can pick one row, and after upgrading:

- **The whole table disappears** when it's wrapped in a `nys-radiogroup`.
- **The radio buttons are invisible** (though still in the page) if the table has radio buttons without a group around them.
- **Radio groups that used to sit side-by-side (horizontal) are now forced to stack vertically**, with no way to change it.

This is a **breaking change that went out in a routine minor update**, so people hit it without warning.

## Why it happens

In 1.19.0 we reworked radio buttons to fix a real screen-reader bug (#1553). That rework was correct in its goal, but it changed *who draws the radio buttons*:

- **Before:** each `nys-radiobutton` drew itself, so you could put it anywhere — including a table cell — and it showed up where you placed it.
- **After:** the individual `nys-radiobutton` draws **nothing**. The parent `nys-radiogroup` now looks at its children, throws away everything else you put inside it, and re-draws the radios itself as a single vertical list.

So any table (or other markup) placed inside the group gets hidden, and a radio button on its own no longer shows anything.

The important realization: **putting a radio button in each row of a table is a valid, common pattern** ("pick one row"). It shouldn't be swapped for a checkbox (a checkbox means "toggle each item on/off independently," which is a different thing and can't enforce "pick exactly one"). The component just needs to support the pattern again.

## Next steps

The team reviewed this and decided on the following fix. This mirrors how **`nys-checkbox` already works today** (the checkbox never got this rework, so a checkbox in a table still works — this brings radio back in line with it):

1. **Let `nys-radiobutton` and `nys-checkbox` draw their own button + label again.** The group's job goes back to *wrapping* the buttons, not *creating* them. This alone fixes the disappearing table, the invisible radios, and the forced-vertical layout (you control the layout again).

[...]

PR #1765, opened 2026-07-14
feat(a11y): cross-shadow label association + fix inert error association

plasticmind:
Follow-up to #1632. Wires the (previously unused) `@nysds/internals` ARIA helper into form controls, adds opt-in external light-DOM labelling, and fixes an error-association bug the work surfaced.

Base is `fix/accessibility-remediation`, not `develop` — this builds on the ElementInternals layer in #1632.

## What's here

**Cross-shadow labelling.** New `associateControlRefs()` in `@nysds/internals` sets a control's own `ariaLabelledByElements` (plus a string `aria-label` fallback), letting a control borrow its accessible name from an element **outside** its shadow root — which IDREF attributes cannot reach. `nys-checkbox` uses it via two new props:

- `labelledby` — id of a light-DOM element to take the name from (e.g. a table column `<th>`)
- `hideLabel` — suppress the internal visible `<nys-label>`

This unblocks the checkbox-in-a-table-row pattern. `nys-radiobutton` reuses the same pattern once #1721's render-revert lands; it is out of scope here.

**Nameable label/error.** `<nys-label>` and `<nys-errormessage>` now expose their text via `ElementInternals.ariaLabel`, so references to them resolve without depending on each engine flattening nested-shadow text.

**Error association fix (the significant one).** Chromium **never surfaces `aria-errormessage` for a control inside a shadow root** — true for the IDREF attribute *and* for `ariaErrorMessageElements` reflection. Every NYSDS form control renders its error inside its shadow root, so the error was never associated with its field. The text was still *announced*, because `<nys-errormessage>` renders a `role="alert"` live region — which is exactly why this went unnoticed. But a screen-reader user interrogating an invalid field heard nothing about why.

`aria-describedby` **does** resolve inside a shadow root. Seven controls now reference the error element with both relations — `aria-errormessage` for engines that honor it, `aria-describedby` for the one that actually reaches the AX tree: textinput, textarea, select, datepicker, combobox, radiogroup, checkbox.

Two also gained a missing `aria-invalid`, which Blink requires before exposing any error relation at all:
- `nys-checkbox` — its only `aria-invalid` was on the separate "other" text field
- `nys-combobox` — had none

Also drops `nys-checkbox`'s `aria-describedby="group-info"`, a dangling IDREF pointing at an element that exists nowhere in the codebase.

## Verification

`npm run verify:a11y-names` (new) reads Blink's **real** accessible-name computation over CDP — the only automated way to check this, since browsers expose no computed-name API to page scripts. It asserts three cases against the shipped components:

```
PASS  internal-label   nys-textinput named by <nys-label> (text two shadow roots deep)
PASS  internal-error   error reaches the AX tree as the input's description
PASS  external-th      nys-checkbox named by a light-DOM element across the shadow boundary
```

Full suite: 1028 passing, 34/34 files, branches 81.23%.

This verification is not ceremonial — it caught two bugs that typechecked cleanly and that unit tests alone would have missed: an early version of the labelling wiring stripped the accessible name off *every* checkbox (element-reference reflection clears the content attribute when set to `null`), and the shadow-DOM `aria-errormessage` failure above.

## Not addressed

Filed separately rather than fixed here, since both need a design decision rather than a mechanical change:

- #1763 — `nys-fileinput`'s error is bound to a `hidden`/`aria-hidden` input that isn't the exposed control
- #1764 — `nys-fileitem` passes error *text* where `aria-errormessage` expects an *id*

## Still open

Manual Safari/VoiceOver + Firefox/NVDA verification. Element reflection is Chromium-only today (other engines get the string `aria-label` fallback), and the `aria-describedby` fix is verified only in Blink. Headless engines can't report computed names for WebKit/Gecko, so this needs a real AT pass before we claim cross-engine support.

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -2692,8 +2692,7 @@
             {
               "title": "Inverted",
               "code": "<nys-button label=\"Inverted\" inverted></nys-button>",
-              "lang": "html",
-              "render": "<div style=\"background: #1b1b1b; padding: 1rem;\">\n<nys-button label=\"Inverted\" inverted></nys-button>\n</div>"
+              "lang": "html"
             },
             {
               "title": "Disabled",
@@ -2959,6 +2958,26 @@
             },
             {
               "kind": "field",
+              "name": "labelledby",
+              "type": {
+                "text": "string"
+              },
+              "default": "\"\"",
+              "description": "Id of an element in the host's light-DOM tree to borrow the accessible name\nfrom (e.g. a table column `<th>`). Enables labelling a checkbox that has no\nvisible label of its own.",
+              "attribute": "labelledby"
+            },
+            {
+              "kind": "field",
+              "name": "hideLabel",
+              "type": {
+                "text": "boolean"
+              },
+              "default": "false",
+              "description": "Suppress the internal visible `<nys-label>` (use with `labelledby` for\ntable cells).",
+              "attribute": "hideLabel"
+            },
+            {
+              "kind": "field",
               "name": "_mobileQuery",
               "privacy": "private"
             },
@@ -2985,6 +3004,12 @@
                   "text": "Promise<HTMLInputElement | null>"
                 }
               }
+            },
+            {
+              "kind": "method",
+              "name": "_syncExternalLabel",
+              "privacy": "private",
+              "description": "Point the native <input> at a light-DOM element that IDREF attributes cannot\nreach across the shadow boundary. associateControlRefs sets the control's own\nariaLabelledByElements (honored in Chromium) plus a string aria-label fallback\nfor engines that do not yet resolve element references.\n\nOnly ever called with an external labelledby set: with no external target the\nnative input keeps the internal same-root aria-labelledby IDREF that render()\nemits, and must not be touched."
             },
             {
               "kind": "method",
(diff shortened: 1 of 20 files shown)


plasticmind:
## Manual AT verification — done

Walked the checklist (`docs/accessibility-label-association-manual-verification.md`) against a demo page exercising all six cases.

- **Safari + VoiceOver (macOS)** — working. The table checkbox is announced by its column header, and errors are conveyed on the invalid field.
- **Chrome** — working (also verified programmatically via `npm run verify:a11y-names`, which reads Blink's real AX tree).
- **Firefox** — working.

This closes the last open item in the PR description. One caveat worth stating plainly: we confirmed *that* VoiceOver names the checkbox correctly, but did not record *which mechanism* produced the name — the `ariaLabelledByElements` element reference (if WebKit has shipped it) or the string `aria-label` fallback that `associateControlRefs()` also writes. Both are passes by design, and the user-facing outcome is identical. If we later want to drop the fallback, that distinction needs a deliberate re-test.

## One more bug, caught by the browser

Manual verification also surfaced a rendering bug that **every automated check missed** (5adad455):

With `hideLabel` set, the label guard `!this.hideLabel && ... && html\`...\`` evaluated to the boolean `false`, and lit-html renders `false` as the literal text **"false"** next to every checkbox. The original guard couldn't hit this because it short-circuited on `this.label` — an empty *string*, which lit renders as blank.

The unit tests, the axe audit, TypeScript, and the CDP name verification all passed, because each asserted the *absence* of `<nys-label>` — which was true either way. None looked at what was actually on screen. Fixed with lit's `nothing` sentinel, plus a regression test that asserts on rendered text rather than element absence.

Checkbox suite: 50 passing, branches 81.6%.

PR #1772, opened 2026-07-17
Fix: Remove cloning from `nys-globalheader`

### LINKED ISSUE #1726: FIX: Global header - clones nav links into shadow DOM and forces pathname-based active state, ignoring author `aria-current`

NovaCat35:
# Summary
Remove slot cloning, preserve user/dev provided `aria-current`, and update active link handling to respect application routing. Styling and whatnot should mirror exactly the previous version if I did this right.

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1726 🎟️

## Testing
Done on the [React Demo](https://github.com/ITS-HCD/nysds-react-demo)


esteinborn:
Cloned repo, ran on vanilla demo which loads nysds-full.min.css which overrides the global `a` css rules. This is something to be acutely aware of when writing these lightDOM styles. Hover state makes the links blue text:
image


In fact, we need to have a new demo that loads in multiple css stylesheets from other frameworks to check to make sure they aren't clobbering our lightDOM styles.

esteinborn (changes requested):
Still a problem with styles bleeding through for :active

Image

PR #1773, opened 2026-07-17
Fix: radiobutton rework

### LINKED ISSUE #1721: FIX: nys-radiogroup 1.19 rework broke radios in tables & custom layouts (agreed fix inside)

NovaCat35:
# Summary
Fixes an issue introduced in v1.19.0 where standalone radio buttons (those not inside a radio group) rendered nothing after the component was changed to read only. Standalone radio buttons now render as native <input type="radio"> elements. Also fixes the description slot not rendering.

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1721 🎟️

## Testing and review

Testing done in Storybook and on the [React Demo](https://github.com/ITS-HCD/nysds-react-demo)

### Accessibility testing

- [x] Screen reader testing
- [x] Keyboard navigation


packages/nys-radiobutton/src/nys-radiobutton.stories.ts
```diff
@@ -422,3 +517,108 @@ export const Preselected: Story = {
...
+      <td>
+      <label>Brooklyn</label>
+      <input type="radio" name="group" value="Brooklyn"></input>
+      </td>
+      </tr>
+      <tr>
+      <td>Medicaid</td>
+      <td>
+      <label>Manhattan</label>
+      <input type="radio" name="group" value="Manhattan"></input>
```
mendeza:
```suggestion
      <label for="rg-bk1">Brooklyn</label>
      <input type="radio" name="group" value="Brooklyn" id="rg-bk1"></input>
      </td>
      </tr>
      <tr>
      <td>Medicaid</td>
      <td>
      <label for="rg-bk2">Manhattan</label>
      <input type="radio" name="group" value="Manhattan" id="rg-bk2"></input>
```

esteinborn:
PR has a merge conflict that needs to be resolved before I can review. @NovaCat35

mendeza:
Current behavior is good; reliance on `aria-label` and `aria-hidden` is rather overwrought, but unavoidable with current architecture; Robert says there is more revision coming at the group level, will re-test when available.

PR #1781, opened 2026-07-23
Fix: read component dependencies from package.json 

### LINKED ISSUE #1760: FIX: `npm run gen-stories` doesn't account for dependencies
## Summary
The new build process for storybook and and running script `npm run gen-stories` doesn't generate dependencies like if `nys-alert` or vertical nav imports in `nys-icon` as a dependency

emilygorelik:
Before the gen-stories script juct checked what components were used in the examples to figure out what needs to be imported to storybook. Now it also reads the package.json file. This mostly effected icons when they were not directly used in a component but needed to be rendered. In the below example the `nys-button` is read by the gen-stories script and knows to import button but it couldn't tell that icon needed to be imported too since it was used as a prop.

```
<nys-textinput id="search-demo" type="search" placeholder="Search">
  <nys-button
    slot="endButton"
    ariaLabel="Search"
    prefixIcon="search"
    onclick="alert('searching for: ' + document.getElementById('search-demo').value)"
  ></nys-button>
</nys-textinput>
```

### DIFF
--- a/packages/nys-accordion/src/nys-accordion.stories.ts
+++ b/packages/nys-accordion/src/nys-accordion.stories.ts
@@ -2,6 +2,7 @@ import { html } from "lit";
 import { Meta, StoryObj } from "@storybook/web-components-vite";
 import "./nys-accordion";
 import "./nys-accordionitem";
+import "@nysds/nys-icon";
 
 const meta: Meta = {
   title: "Components/Accordion",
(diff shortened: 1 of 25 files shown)


NovaCat35 (approved):
LGTM

esteinborn (approved):
LGTM;

PR #1792, opened 2026-07-24
feat: issue#1791, shx added as a dev-dependency, 'cem' script updated…

### LINKED ISSUE #1791: Enable terminal environments without coreutils commands like `cp` to run npm scripts
When running the npm script  `npm run cem` or any other script that uses that script (i.e. `npm run gen-stories`) in a terminal environment that doesn't recognize coreutil commands, the script won't be fully executed.

P.S. I'm assuming the scripts that run `npm run cem` are still being used. I had no specific reason for running these scripts other than testing them out to see what they do.

## Steps to reproduce
- Start a PS session from the project's root directory
- Run the command `npm run cem`
- Verify script can't be fully executed because `'cp' is not recognized...`

Image

PS Version: 5.1.26100.8655

## Request
While a Windows-based Coreutils library exists and can be installed by the user, it would be nice for these scripts to work "out of the box" without having to change shells.

One potential solution is https://github.com/shelljs/shx. If this is included in the project as a dev-dependency, references to commands like `cp` in the package.json scripts can be replaced by the shx equivalent, and allow for PS users to run `npm run cem` without installing a separate Coreutils library.

Hiroshi-nobuoka2-dolbas:
… to use shx cp

# Summary

https://www.npmjs.com/package/shx package added as a dev dependency so that coreutils commands like `cp` can be executed in shells that don't normally recognize them. 

The "cem" script in package.json uses the `cp` command, and would fail if the script were run in "base" Powershell. 

Adding this dependency allows for devs using PS to run scripts like `npm run cem` or `npm run gen-stories` without installing a separate library (they just have to run `npm install` beforehand, as they normally would upon the initial installation of the project).

## Breaking change

_Indicate if this update is a breaking change with **one** of the following statements:_

## Related issues

Closes #1791 

## Problem statement

_What problem does this solve? Who is impacted? Why is it important?_

Some Windows users use Powershell, which doesn't recognize coreutils commands by default. The desired state is to be able to run commands in the package.json scripts regardless of the shell the script is executed in.

## Solution

Added the shx package as a dev dependency. I haven't explored other solutions but this solution prevents the need to have separate scripts in package.json to accommodate different shell environments.

After update, in Powershell:

image

After update, in Git Bash (windows):

image

## Testing and review

_Provide testing instructions and any special notes for the reviewer._

To verify scripts run in Powershell:
```powershell
# Open a Powershell session from the project's root directory, then run
npm install

# Remove file that gets copied if it exists. (rm is a built-in alias for Powershell, unlike cp)
rm .\dist\custom-elements.json

# Verify this runs and that the file has been copied over
npm run cem; Test-Path -Path .\dist\custom-elements.json
```

To verify scripts still work in bash/zsh/etc:
```bash
# Open a terminal session from the project's root directory, then run
npm install

# Remove file that gets copied if it exists.
rm dist/custom-elements.json

# Verify this runs and that the file has been copied over
 npm run cem && [ -f "./dist/custom-elements.json" ] && echo "File exists."
```

## Additional considerations

_Anything else to call out? (performance, dependencies, migration notes, backward compatibility, security)_

`package-lock.json` gets updated after adding this dependency and removes the `"peer": True ` field in several existing packages, which I think means they've become more explicit dependencies. Wasn't sure if this was desired so I didn't include this update in the commit.

## Dependency updates
| Dependency name              | Previous version | New version |
| ---------------------------- | :--------------: | :---------: |
|shx     |        --        |   [0.4.0]   |

## Before submitting this PR, confirm:
- [ ] The code follows the NYSDS coding guidelines.
- [ ] Accessibility checks are complete (see above).
- [x] Run `git pull origin [base branch]` to pull the latest from the base branch and resolve conflicts (our base branch is `develop`).
- [ ] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [x] Run `npm run test` and confirm that all tests pass.
- [ ] Update relevant documentation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,6 +52,7 @@
         "rollup-plugin-minify-template-literals": "^1.1.7",
         "rollup-plugin-visualizer": "^6.0.5",
         "sass": "^1.94.0",
+        "shx": "^0.4.0",
         "sinon": "^21.0.0",
         "storybook": "10.3.5",
         "stylelint": "^17.1.1",
@@ -427,9 +428,9 @@
       }
     },
     "node_modules/@emnapi/wasi-threads": {
-      "version": "1.2.2",
-      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.2.tgz",
-      "integrity": "sha512-c95qOXkHdydNKhscBTebqEC1CVAZpyqOfVfBzQ1qgzyl3gfeldUjIggDbIZgDKsHLgnsM+igH7TJ/eAasaVuMA==",
+      "version": "1.2.3",
+      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.3.tgz",
+      "integrity": "sha512-ELEBe8PsLvvJ6QMr0zLt8ffvOHW/dc1m3CEzNMg7aJUv3bMaoDtw2TXyDAwkYBuroxxuHEwhRTLJSe5sya547g==",
       "dev": true,
       "license": "MIT",
       "optional": true,
@@ -9309,6 +9310,13 @@
         "node": ">= 0.4.0"
       }
     },
+    "node_modules/nice-try": {
+      "version": "1.0.5",
+      "resolved": "https://registry.npmjs.org/nice-try/-/nice-try-1.0.5.tgz",
+      "integrity": "sha512-1nh45deeb5olNY7eX82BkPO7SSxR5SSYJiPTrTdFUVYwAl8CKMA5N9PjTYkHiRjisVcxcQ1HXdLhx2qxxJzLNQ==",
+      "dev": true,
+      "license": "MIT"
+    },
     "node_modules/no-case": {
       "version": "3.0.4",
       "dev": true,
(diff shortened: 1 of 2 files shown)


NovaCat35 (dismissed):
LGTM!

esteinborn:
@Hiroshi-nobuoka2-dolbas Thank you for this contribution. In fact, you are the first person outside of the Design System team to submit a PR! YOU ARE A LEGEND!

In order to ensure you don't remove those `peer: true`, make sure you are on the latest LTS version of node.js: 24.13.x (latest is 24.18.3). this should stop the removal of those `peer` settings when you run a new `npm i`.

Reason: we need the updated package-lock.json for this to run on our CI/Cd pipeline.

Hiroshi-nobuoka2-dolbas:
@NovaCat35 @esteinborn Thank you!! 

I updated my node version to the latest LTS 24.18.X version and the npm version to what you have in your `package.json` under the ["packageManager"](https://github.com/ITS-HCD/nysds/blob/main/package.json#L80) field. 

I tried running `npm i`  on the `develop` git branch to test for any changes to the lock file. If I'm using the same node/npm versions I assumed the generated lock file would match what's on Github.

One package received a patch update (diff is pasted below). Changing from node v24.18.1 to v24.13.0 resulted in the same changes to the lock file, so i thnk it's just the npm version that affects it. 

I pushed the lock file update with my added dependency as a new commit, the peer updates are gone as prophesized! I'll try to create another ticket for enforcing consistent node versions.

---

node v24.18.1 (LTS)
npm v11.6.2

Running `npm i --allow-git=all` on the `develop` branch after deleting `node_modules` and `package-lock.json`:
```diff
diff --git a/package-lock.json b/package-lock.json
index c35effb5..e14a5b41 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -427,9 +427,9 @@
       }
     },
     "node_modules/@emnapi/wasi-threads": {
-      "version": "1.2.2",
-      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.2.tgz",
-      "integrity": "sha512-c95qOXkHdydNKhscBTebqEC1CVAZpyqOfVfBzQ1qgzyl3gfeldUjIggDbIZgDKsHLgnsM+igH7TJ/eAasaVuMA==",
+      "version": "1.2.3",
+      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.3.tgz",
+      "integrity": "sha512-ELEBe8PsLvvJ6QMr0zLt8ffvOHW/dc1m3CEzNMg7aJUv3bMaoDtw2TXyDAwkYBuroxxuHEwhRTLJSe5sya547g==",
       "dev": true,
       "license": "MIT",
       "optional": true,
```

esteinborn:
@Hiroshi-nobuoka2-dolbas After reviewing this, it seems like it works fine. But upon further investigation, there are still calls to the `cp` command in the repo. So I'm interested in seeing how your `npm run build:all` script is running, and if it's failing at the last step — `mcp-server:build` (which it should since it's running plain old `cp`)

`cp` calls exist in: `/packages/mcp-server/package.json:copy-data` script

Let us know what your tests find!

Hiroshi-nobuoka2-dolbas:
@esteinborn Thank you for reviewing! Yes I ran into several issues when running `npm run build:all`. I'm now worried my proposed strategy can lead to the same scripts yielding different results, and other "it works on my machine"-like issues. 

At this point I think it would be safer to just have Windows users use some nix shell derivative, most should have something like Git Bash already installed. There's also an npm config `script-shell` that can be set to a shell executable, so with the command below I was able to fully execute the `npm run build:all` command.

```
npm config set script-shell <filepath for bash.exe>
```

It also occurred to me that Windows Subsystem for Linux is an available development environment for Windows-users, where bash is usually the default shell.

I'm totally good with closing the PR without merging, unless you believe it's worth continuing. An alternate contrib I can make is adding a section to the wiki https://github.com/ITS-HCD/nysds/wiki#setting-up-locally? There's already a note that mentions installing OpenSSL on Windows, so I can create a subsection for Windows installation/setup which includes that and anything else they'll need to run the npm scripts successfully.

esteinborn:
@Hiroshi-nobuoka2-dolbas Great detailed breakdown of the problem and possible solutions. I think you are right about limiting the customization of the package to accommodate windows Powershell/CMD environments. Users must install Git on their windows machines in order to download the nysds repo, and would then be able to run all scripts for it inside Git Bash (as you indicated). 

In this case, lets lean on the included tools with Git, rather than a workaround for windows powershell/cmd users. 

Last note: The scripts you have been running are the scripts used by internal NYSDS team developers to create the @nysds/components NPM package. Typically, unless a user is contributing back, they wont need to run these commands. 

P.P.S. Thanks for the first PR to NYSDS!

PR #1796, opened 2026-07-27
ENHANCEMENT: button a11y improvements

### LINKED ISSUE #1464: Button improvement requests
# Overview

- The Button component has attribute `tabindex="0"`.
  - **Request**: this is natively provided and can be removed.
- The Button component, in its default mode as a 'button', has `role="button"`.
  - **Request**: this is redundant and can be removed. 
- The Button component renders `aria-label=<labelvalue>` and doesn't allow authors to set an arbitrary `aria-label` value. The content of `.nys-button__text` is the accessible name and recognized by ATs. Thus the `aria-label`, which overrides it, is redundant and heavy-handed. 
  - Consideration: a method exists to work around this when a `circle` button has only an icon as visible content. In the assumption that the team does not wish authors to make full-size icon buttons, this still leaves the scenario when additional context is needed for ATs.
  - **Request**: with `circle` button, keep the labeltext but make it __visually-hidden__, e.g.,  `<div class="nys-button__text sr-only">labeltext</div>`
  - **Request**: Allow arbitrary user-provided `aria-label` values to be passed down from the WC to the shadow-root. And while at it, `aria-describedby` also.

## Annotations

- **`[A]`** &rarr; failure of ARIA association
- **`[B]`** &rarr; it makes sense that the WC attribute `label` maps to the accessible text content of the rendered HTML
- **`[C]`** &rarr; `<button>` doesn't need `role="button"`
- **`[D]`** &rarr; `tabindex="0"` is already a native behavior for these elements
- **`[E]`** &rarr; `aria-label` values rendered in the shadow-root ignore authors' legitimate need to set arbitrarily

Annotated screenshot of buttons and their rendered code showing poor use of attributes

emilygorelik:


esteinborn (dismissed):
Do we have a way to test out these changes? @emilygorelik

packages/nys-button/src/nys-button.ts
```diff
@@ -520,8 +526,8 @@ export class NysButton extends LitElement {
                   this.suffixIcon ||
                   "button",
               )}
+              aria-describedby=${ifDefined(this.ariaDescribedBy || undefined)}
               aria-description=${ifDefined(this.ariaDescription || undefined)}
-              role="button"
```
mendeza, line 524: Two points.
Please remove all uses of `aria-label`. Since `sr-only` is already employed for the Circle variant, `aria-label` can be safely deleted. Then the `label` prop would be made required (in all cases but particularly for Circle).

Please remove all uses of `aria-description`. It has poor support and what it might be used for is better covered by `aria-describedby`.

mendeza (changes requested):

esteinborn (approved):
LGTM;

PR #1802, opened 2026-07-28
Fix: `nys-dropdown` to toggle via keyboard once 

### LINKED ISSUE #1800: BUG: dropdown button inside of `nys-globalheader` no longer toggles via keyboard
## Summary
Bug found on inside `nys-globalheader` where a `nys-button` use to trigger `nys-dropdown` is no longer working. Need to fix issue and add regression test.

[Troubleshooting Thread Link](https://teams.microsoft.com/l/message/19:0228156e2bb5419c8152047f596a7bfb@thread.tacv2/1785243492279?tenantId=f46cb8ea-7900-4d10-8ceb-80e8c1c81ee7&groupId=40dc9e8f-13b9-4301-8cb7-db5c2b21c9fa&parentMessageId=1785243492279&teamName=NYS%20Design%20System&channelName=Troubleshooting&createdTime=1785243492279)
> **Title:** nys-globalheader button does not toggle nys-dropdownmenu on keyboard input
> **Developer:** Brislin, Jess (ITS)
> 
> I noticed that when using an nys-button inside nys-globalheader to trigger an nys-dropdownmenu, a mouse click will correctly toggle the dropdown but keyboard input will not. This is on 1.19.3. On our other application using the same header and dropdown code, which is a few versions behind on 1.18.3, keyboard input still works, so I believe this might be a regression.
> _[See attached video on troubleshooting thread]_

NovaCat35:
# Summary
> [!NOTE]
> I spend more time crafting out the test that keeps failing than the actual solution itself...I'm gonna cry 😭

Fix issue with `nys-dropdown` triggering the toggle state twice due to `nys-button`'s click event and the dropdown's keydown listener both calling the keydown event.

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1800 🎟️

## Testing 
Done on storybook and the React Demo


packages/nys-dropdownmenu/src/nys-dropdownmenu.ts
```diff
@@ -51,6 +51,31 @@ interface SpaceAvailable {
  *   </nys-dropdownmenu>
  * </div>
  * ```
+ *
+ *  * @example Test
```
esteinborn: extra asterisk before @example

esteinborn (changes requested):
one ver ysmall change and its GTG

PR #1807, opened 2026-07-31
FIX: scheduled an update

### LINKED ISSUE #1555: FIX: Scheduled an Update
Correct the following warnings:
```
packages/nys-combobox/src/nys-combobox.test.ts:

 🚧 Browser logs:
      Element nys-combobox scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update. See https://lit.dev/msg/change-in-update for more information.

packages/nys-select/src/nys-select.test.ts:

 🚧 Browser logs:
      Element nys-select scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update. See https://lit.dev/msg/change-in-update for more information.

packages/nys-radiobutton/src/nys-radiobutton.test.ts:

 🚧 Browser logs:
      Element nys-radiobutton scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update. See https://lit.dev/msg/change-in-update for more information.

NOT EXAUSTIVE

emilygorelik:
Also updated template and docs so we avoid this in the future

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -58,6 +58,33 @@ Each package: `tsc --emitDeclarationOnly` → `vite build`. Package vite configs
 - CI runs on PRs to develop/main
 - `nys-label` is for form components only. Elsewhere, render label/description text as plain markup in the component's own shadow DOM.
 
+### Lifecycle: derive state in `willUpdate`
+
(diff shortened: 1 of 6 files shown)


esteinborn (approved):
LGTM;

PR #1863, opened 2026-08-26
Enhancement: Unavheader translate defaults to localize

### LINKED ISSUE #1857: ENHANCEMENT: UNav Header Localize support by default
Unav header needs to transition to localize support as the default translation option providing a deprecated smartling fallback for sites who have not made the transition.

Think:
```
<nys-unavheader> <-- Default shows translate dropdown with 12 languages, loads and uses localize.

<nys-unavheader hideTranslate> <-- Hides the translate button and does not load localize.js script

<nys-unavheader smartling> <-- Shows the old, subdomain-based, switching method which works with Smartling.
```

### Development
- Make localize default translation when enabled
- Only load localize JS if hideTranslate === false || smartling === true
- OPTIONAL: smartling as dedicated property
- auto-list 12 languages by default
- autoapprove enabled by default
- load disclaimer on translate (except English)
	- All disclaimers implemented
	- notranslate property on all disclaimers
- RTL trigger (do we use CSS ) (test out performance of css vs js attribute)
```
html[lang="ur"],
html[lang="yi"] {
	direction: "rtl";
}
```
- Bundle unav into a single file
	- custom vite script that bundles all the dependencies in one file
**Development Qs:**
Q: do we create a new unav bundle, or hijack the old one.
	currently: globa-bundle.js is loaded and we replace the div with unav code
	Proposal:
		fully replace the global-bundle with NYSDS version, no fallback
		Create a new script URL with NYSDS unav and reference that in the new code
		NYSDS version of unav, but also checks for <div id="unav-header"></div>
**Q:** do we persist or track disclaimer display with a cookie? (if closed, don't re-open etc)
	What do they do right now???
	
### Test
- Deploy unav bundle to webNY DEV endpoint (cloudflareCDN DEV)
- test integration with Drupal ACSF
- test integration with Drupal Single Instance
- test integration with static
- test integration with a react/angular/next/vue/.net/java app
### Communication
- develop migration guide from previous to new
- Set migration date
- be clear that unav will no longer function unless you update the codebase for it
### Deploy
- Deploy unav bundle to webNY Prod endpoint (CloudflareDN PROD)
- Update source code on sites

emilygorelik:
- new `translateKey` for passing in an ID
- currently if no `translateKey` is passed in enable `hideTranslate` but I can change this to use smartling instead
- added tests for new functionality


esteinborn:
Fixed subdomain trigger (might have been working before, but it definitely works now)
Fixed disclaimer appearing on all translated pages, now it only activates on click of a language.

PR #1869, opened 2026-08-28
FIX: Combobox form required when not explicitly set

### LINKED ISSUE #1596: BUG: combobox doesn't focus on component for invalid submission
## Summary
- [ ] Component `nys-combobox` doesn't focus on the component for invalid form submission. The expected output is that you hit the submit button and it scrolls back to the component. 
- [ ] When the combobox is marked as "required", unfocusing from the select without choosing an option makes the require validation disappear
- [ ] Combobox, when not "required," shows an error state saying the field needs to be filled when I autofill (but not select an option) and unfocus

All 3 needs fixing and testing for form submission to work properly.

## Screenshot

Image

emilygorelik:
Combobox was getting marked as required when not explicitly set forcing a selection. Fixed

### DIFF
--- a/packages/nys-combobox/src/nys-combobox.test.ts
+++ b/packages/nys-combobox/src/nys-combobox.test.ts
@@ -221,6 +221,22 @@ describe("nys-combobox", () => {
     expect(input?.hasAttribute("required")).to.be.false;
   });
 
+  it("does not trigger validation error on blur when required is not set and value is empty", async () => {
+    const el = await fixture<NysCombobox>(html`
+      <nys-combobox>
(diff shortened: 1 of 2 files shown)


NovaCat35 (dismissed):
Validation issue still persists when "required" is set and the user autofills and unfocuses. This bypasses requirement validation.

## Recorded video on this issue
https://teams.microsoft.com/l/message/19:b7d785298d77407681720a48d609d058@thread.v2/1787927672469?context=%7B%22contextType%22%3A%22chat%22%7D

## Related ticket
https://github.com/ITS-HCD/nysds/issues/1596 🎟️

PR #1872, opened 2026-08-28
Fix: issue with duplicate value on radiobutton with form submission

### LINKED ISSUE #1871: BUG: `nys-radiobutton` and `nys-radiogroup` submits value multiple time with form.
## Summary
The radio buttons submit the checked value multiple times in standard HTML forms, resulting in duplicate values like ["weekly", "weekly", "weekly"]. This happens with both <nys-radiogroup> and manually grouped <nys-radiobutton> components using the name attribute.

NovaCat35:
# Summary
Fix issue with `nys-radiobutton` submitting the checked value multiple times in standard HTML forms, resulting in duplicate values like ["weekly", "weekly", "weekly"]. This happens with both and manually grouped components using the name attribute.

## Breaking change

## Related issues

Closes https://github.com/ITS-HCD/nysds/issues/1871 🎟️

## Testing
Tested on the React Demo

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -4,6 +4,834 @@
   "modules": [
     {
       "kind": "javascript-module",
+      "path": "packages/angular/node_modules/@babel/core/src/config/files/index-browser.ts",
+      "declarations": [
+        {
+          "kind": "function",
+          "name": "findConfigUpwards",
+          "return": {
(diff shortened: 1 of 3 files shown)


esteinborn (approved):
LGTM;

